PR #1, opened 2017-06-21
one estimation directory capability

dpastoor:
This is to handle not re-running a model if submitted and detects and output directory already. Can be handy when submitting large directories of models and only want to re-run the ones that were removed or newly added.


dpastoor:
logic seems sound and prelim tests work

PR #11, opened 2019-07-26
wip: add new struct types to support summary data

david-lyder:


parsers/nmparser/parse_run_details.go
```diff
@@ -71,5 +60,14 @@ func ParseRunDetails(lines []string) RunDetails {
 		covarianceTime,
 		functionEvaluations,
 		significantDigits,
+		// TODO: populate with values
```
dpastoor: now that rundetails is more complex, we should not positionally match, and instead should explicitly refer to values 

```
RunDetails{
ProblemText: ""
...
}
```

parsers/nmparser/parse_run_details.go
```diff
@@ -71,5 +60,14 @@ func ParseRunDetails(lines []string) RunDetails {
...
+		"",  // ProblemText 		string
+		"",  // ModFile             string
+		nil, // EstimationMethod    []string
```
dpastoor: should probably be an empty []string not nil right?

parsers/nmparser/structs.go
```diff
@@ -0,0 +1,65 @@
...
+type CovarianceStep struct {
+	Attempted  bool
+	MatrixType string
+	OK         bool
+	Errors     []string
+}
+
+// Shrinkage ...
+type Shrinkage struct {
+	SD []float64
```
dpastoor: wasn't there one more shrinkage element?

david-lyder: We had 3 structures defined,  but the type looked the same for all, so I create a single type and 3 instances within the ShrinkageDetails type. The original 3 were:

EtaShrinkage {
SD: []float64
VR: []float64
}
EbvShrinkage {
SD: []float64
VR: []float64
}
EpsShrinkage {
SD: []float64
VR: []float64
}

parsers/nmparser/parse_run_details.go
david-lyder: ok, changed

parsers/nmparser/parse_run_details.go
david-lyder: sure, I originally saw this as throw-way code to satisfy the compiler, but now realize it may stick around for a while ... change is in progress ...

Issue #8, opened 2019-07-25
refine output summary


dpastoor:
parsing a model results should surface at least the following:

* [ ] Parameter table
* [ ] Nonmem version
* [ ] Estimation method
* [ ] Whether/which Multiple methods
* [ ] Output tables
* [ ] SEE
* [ ] shrinkage
* [ ] OFV
* [ ] Dataset used
* [ ] Number of patients
* [ ] Number of observations
* [ ] Completion messages or problems
* [ ] COV step completed
* [ ] 0 gradient
* [ ] Other signalling ...etc

dpastoor:
* what of these are already done (check box them)
* what is incomplete - eg OFV only gives one OFV value, but now the output looks like:

```
 N*LOG(2PI) CONSTANT TO OBJECTIVE FUNCTION:    817.85529455215863     
 OBJECTIVE FUNCTION VALUE WITHOUT CONSTANT:    723.60167739548945     
 OBJECTIVE FUNCTION VALUE WITH CONSTANT:       1541.4569719476481     
 REPORTED OBJECTIVE FUNCTION DOES NOT CONTAIN CONSTANT
```

* what doesn't exist

dpastoor:
Existing tree for a simple model looks like:

```
{
	"RunDetails": {
		"NMversion": "7.4.3",
		"RunStart": "",
		"RunEnd": "",
		"EstimationTime": 0,
		"CovarianceTime": 0,
		"FunctionEvaluations": 176,
		"SignificantDigits": 3.2
	},
	"FinalParameterEstimates": {
		"Theta": [
			1.82,
			12.9,
			0.383
		],
		"Omega": [
			0.0595,
			0,
			0.0266,
			0,
			0,
			0.302
		],
		"Sigma": [
			0.0365,
			0,
			0.116
		],
		"OmegaCorr": [
			0.244,
			0,
			0.163,
			0,
			0,
			0.55
		],
		"SigmaCorr": [
			0.191,
			0,
			0.341
		]
	},
	"FinalParameterStdErr": {
		"Theta": [
			0.0692,
			0.652,
			0.0361
		],
		"Omega": [
			0.0138,
			0,
			0.0212,
			0,
			0,
			0.0724
		],
		"Sigma": [
			0.00419,
			0,
			0.0337
		],
		"OmegaCorr": [
			0.0284,
			0,
			0.0651,
			0,
			0,
			0.0659
		],
		"SigmaCorr": [
			0.011,
			0,
[...]

PR #12, opened 2019-07-30
wip: support first set of new summary data. unit test passes, int. te…

david-lyder:
…st pending
#8


parsers/nmparser/parse_run_details.go
```diff
@@ -31,27 +49,62 @@ func ParseRunDetails(lines []string) RunDetails {
...
 			runStart = replaceTrim(line, "Started")
 		case strings.Contains(line, "Finished"):
 			runEnd = replaceTrim(line, "Finished")
+		case strings.Contains(line, "Stop Time:"):
+			if i+1 < len(lines) {
+				runEnd = lines[i+1]
+			}
+		case strings.Contains(line, "$PROB"):
+			problemText = replaceTrim(line, "$PROB")
+		case strings.Contains(line, "$ESTIMATION MAXEVAL"):
```
dpastoor: we should parse this out elsewhere. There are a number of ways this will fail.

user can define shorthand `$EST`, you do not need to do MAXEVAL directly after `$ESTIMATION METHOD=1 MAXEVAL=9999`, etc. To abstract this, we should test for `$EST` then feed that entire line into a parseEstimationDetails type function so we can change that implementation as we go a little more easily

parsers/nmparser/parse_run_details.go
```diff
@@ -31,27 +49,62 @@ func ParseRunDetails(lines []string) RunDetails {
...
+			estimationMethod = append(estimationMethod, parseValue(line, "METH="))
+		case strings.Contains(line, "$DATA"):
+			dataSet = parseLine(line, 1)
+		case strings.Contains(line, "TOT. NO. OF INDIVIDUALS:"):
+			numberOfPatients, _ = strconv.ParseInt(replaceTrim(line, "TOT. NO. OF INDIVIDUALS:"), 10, 64)
+		case strings.Contains(line, "TOT. NO. OF OBS RECS:"):
+			numberOfObs, _ = strconv.ParseInt(replaceTrim(line, "TOT. NO. OF OBS RECS:"), 10, 64)
+		case strings.Contains(line, "NO. OF DATA RECS IN DATA SET:"):
+			numberOfDataRecords, _ = strconv.ParseInt(replaceTrim(line, "NO. OF DATA RECS IN DATA SET:"), 10, 64)
+		case strings.Contains(line, "$TABLE NOPRINT ONEHEADER FILE="):
```
dpastoor: same thing as $EST, this is definitely too specific, we should just grab the entire table line, then parse it elsewhere

testdata/2.lst
```diff
@@ -0,0 +1,473 @@
...
+   DOSE AMOUNT DATA ITEM IS DATA ITEM NO.:   6
+
+0PK SUBROUTINE CALLED WITH EVERY EVENT RECORD.
+ PK SUBROUTINE NOT CALLED AT NONEVENT (ADDITIONAL OR LAGGED) DOSE TIMES.
+0ERROR SUBROUTINE CALLED WITH EVERY EVENT RECORD.
+1
+
+
+ #TBLN:      1
+ #METH: First Order Conditional Estimation with Interaction
```
dpastoor: this could be where we should pull out the estimation method from, instead of trying to hand parse the $EST

parsers/nmparser/parse_lst_file.go
```diff
@@ -25,15 +25,15 @@ func parseOFV(line string) float64 {
...
 // ParseLstEstimationFile parses the lst file
 func ParseLstEstimationFile(lines []string) LstData {
+	ofvIndex := 0
+	startParameterStructuresIndex := 0
+	endParameterStucturesIndex := 0
+	finalParameterEstimatesIndex := 0
+	standardErrorEstimateIndex := 0
+	covarianceMatrixEstimateIndex := 0
+	startThetaIndex := 0
+	endSigmaIndex := 0
```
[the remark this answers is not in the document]
david-lyder: hmm, Ok, then I just need the test before accessing by these indices.  Why would an index ever be a float?

parsers/nmparser/parse_run_details.go
```diff
@@ -22,6 +22,24 @@ func replaceTrim(line string, replacement string) string {
 	return strings.TrimSpace(strings.Replace(line, replacement, "", -1))
 }
 
+func parseValue(line string, value string) string {
+	tokens := strings.Split(line, " ")
```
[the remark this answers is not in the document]
david-lyder: you pass in a value such as "FILE=", or "METH=" and it returns the assigned value.  string.Fields is better

parsers/nmparser/parse_run_details.go
```diff
@@ -31,27 +49,62 @@ func ParseRunDetails(lines []string) RunDetails {
...
+	for i, line := range lines {
 		switch {
 		case strings.Contains(line, "1NONLINEAR MIXED EFFECTS MODEL PROGRAM (NONMEM) VERSION"):
 			nmversion = parseNMVersion(line)
 		case strings.Contains(line, "NO. OF FUNCTION EVALUATIONS USED"):
 			functionEvaluations, _ = strconv.ParseInt(replaceTrim(line, "NO. OF FUNCTION EVALUATIONS USED:"), 10, 64)
 		case strings.Contains(line, "NO. OF SIG. DIGITS IN FINAL EST.:"):
 			significantDigits, _ = strconv.ParseFloat(replaceTrim(line, "NO. OF SIG. DIGITS IN FINAL EST.:"), 64)
-		case strings.Contains(line, "Elapsed estimation time in seconds:"):
+		case strings.Contains(line, "Elapsed estimation"):
```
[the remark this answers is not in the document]
david-lyder: I thought I saw the line of text differently in some of the other lst files, but I just looked and cannot confirm.  The text does appear consistent as:
Elapsed estimation  time in seconds:    19.04
I'll change this back if I don't discover why I did this initially.

parsers/nmparser/parse_run_details.go
david-lyder: OK.

david-lyder:
I committed ddmore.Rmd - I used it for searching through the lst files. The elements that can extend more than one line (and therefore need "chunk" support) are: $INPUT, $EST, $TABLE.
$MODEL is more than one line, but the data is not needed.
METH, #METH, METHOD, etc, required special processing.

To go forward, I propose:
[] Address the feedback on this pull request and merge this branch back to  v1.1.0.  This may include dropping support for some of the values. This branch, feature/summary-1,  was only intended to include trivial values - anything more complicated than a single-line grab can be dropped from this branch and included later.

[] Create a new branch (feature/multiline) to support getting code chunks and passing the chunk to parsing functions with the NM version as a parameter.

[] After merging feature/multiline back to v1.1.0, address the rest of the required values individually, in multiple branches (depending on the complexity of the task) for values such as OBV, Shrinkage, etc.)

[] Concurrently, create unit tests using some of the ddmore files, as well as strings of text from other lst files.

Let me know what you think.

ddmore.Rmd
```diff
@@ -0,0 +1,104 @@
...
+
+```{r}
+## one big df
+
+data_dir <- "./testdata/nmoutputs"
+file_paths <- list.files(data_dir, "\\.lst$", full.names = TRUE)
+bdf <- function(.filename){
+  df <- data.frame( filename = basename(.filename),  lines = readLines(.filename, skipNul = TRUE))
+}
+df_list <- lapply(file_paths, bdf)
```
dpastoor: lapply do.call rbind pattern can be replaced with a single call to map_df()

ddmore.Rmd
```diff
@@ -0,0 +1,104 @@
...
+```
+
+
+```{r}
+## one big df
+
+data_dir <- "./testdata/nmoutputs"
+file_paths <- list.files(data_dir, "\\.lst$", full.names = TRUE)
+bdf <- function(.filename){
+  df <- data.frame( filename = basename(.filename),  lines = readLines(.filename, skipNul = TRUE))
```
dpastoor: readr::read_lines will be faster than readLines by about 10x

ddmore.Rmd
```diff
@@ -0,0 +1,104 @@
...
+```{r}
+## one long list of lines
+lines <- unlist(lapply(file_paths, readLines, skipNul = TRUE)) %>% 
+          grep(pattern = "^[^;]", value = TRUE, useBytes = TRUE) 
+```
+
+
+```{r}
+## distinct list of lines
+dlines <- unlist(lapply(file_paths, readLines, skipNul = TRUE)) %>%
```
dpastoor: you might want to just have all the lines in a single dataframe. Eg the columns would be

file | line_contents | flag1 | flag2 | etc

where flag1/2/3 could be based on various checks you do in each line, so you might have has_dollar_est, has_dollar_table, etc. Then you can do more summaries against that, like

df %>% filter(has_dollar_est) %>% count(file) # would show number of $EST lines per file

dpastoor:
@david-lyder agree on your proposed path forward

parsers/nmparser/parse_run_details.go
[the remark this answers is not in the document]
david-lyder: Found it. The lst files I generated have a extra space:
Elapsed estimation  time in seconds:
                                ^

parsers/nmparser/parse_lst_file.go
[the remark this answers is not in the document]
david-lyder: changed back to var declarations ...

ddmore.Rmd
```diff
@@ -20,10 +20,23 @@ data_dir <- "./testdata/nmoutputs"
 file_paths <- list.files(data_dir, "\\.lst$", full.names = TRUE)
 bdf <- function(.filename){
   df <- data.frame( stringsAsFactors = FALSE,
-                    filename = as.character(basename(.filename)),  
-                    lines = readr::read_lines(.filename)) 
+                    file_name = as.character(basename(.filename)),  
+                    line_text = readr::read_lines(.filename))  %>%    
+                        mutate (line_number = row_number()) %>% 
+                          select(file_name, line_number, line_text)
```
dpastoor: what's the point of this final select? seems these are they only columns anyway? From looking it seems you're actually just trying to reorder the columns so line number is second, not last. If so, one thing you can do so you don't accidentally start dropping columns if you end up adding more later is to do `select(file_name, line_number, everything()`, which will auto add every column after

ddmore.Rmd
```diff
@@ -20,10 +20,23 @@ data_dir <- "./testdata/nmoutputs"
...
 }
-df <- map_df(file_paths[1:2], bdf)
+df1 <- map_df(file_paths, bdf)
+df <- df1 %>% 
+        mutate(has_dollar_est = grepl(x = df$line_text, pattern = "$EST", fixed = TRUE, useBytes = TRUE)) %>%
```
dpastoor: this will still match like $ESTIMATE right? I believe so, just confirming

ddmore.Rmd
```diff
@@ -20,10 +20,23 @@ data_dir <- "./testdata/nmoutputs"
 file_paths <- list.files(data_dir, "\\.lst$", full.names = TRUE)
 bdf <- function(.filename){
   df <- data.frame( stringsAsFactors = FALSE,
-                    filename = as.character(basename(.filename)),
```
dpastoor: can initialize with `tibble::tibble()` instead of data.frame, and you also then don't need stringsAsFactors = FALSE as that is the default for tibbles

ddmore.Rmd
david-lyder: yes

ddmore.Rmd
david-lyder: ok

ddmore.Rmd
david-lyder: ok

david-lyder:
Committed changes to drop TABLE and get METHOD from "#METH"

dpastoor:
@david-lyder are you looking to commit anything else on this scope?

david-lyder:
no, no more commits planned, I consider this one finished.

dpastoor:
great, all tests are currently passing 

```
=== RUN   TestAddingPathLevel
--- PASS: TestAddingPathLevel (0.00s)
=== RUN   TestCleaningThetaBlock01
--- PASS: TestCleaningThetaBlock01 (0.00s)
=== RUN   TestCleaningThetaBlock02
--- PASS: TestCleaningThetaBlock02 (0.00s)
=== RUN   TestCleaningThetaBlock03
--- PASS: TestCleaningThetaBlock03 (0.00s)
=== RUN   TestFormattingThetaBlock
--- PASS: TestFormattingThetaBlock (0.00s)
=== RUN   TestFormattingThetaBlock02
--- PASS: TestFormattingThetaBlock02 (0.00s)
=== RUN   TestParseBlockResults
--- PASS: TestParseBlockResults (0.00s)
=== RUN   TestParseBlockResultsWithDots
--- PASS: TestParseBlockResultsWithDots (0.00s)
=== RUN   TestParseParameterBlock
--- PASS: TestParseParameterBlock (0.00s)
=== RUN   TestParseFinalParameterEstimates
--- PASS: TestParseFinalParameterEstimates (0.00s)
=== RUN   TestParseParameterStructure
--- PASS: TestParseParameterStructure (0.00s)
=== RUN   TestParseParameterStructures
--- PASS: TestParseParameterStructures (0.00s)
=== RUN   TestParseRunDetails
--- PASS: TestParseRunDetails (0.00s)
=== RUN   TestParseRunDetailsFromFile
--- PASS: TestParseRunDetailsFromFile (0.00s)
=== RUN   TestParseThetaResultsBlock
--- PASS: TestParseThetaResultsBlock (0.00s)
=== RUN   TestCreateDiagonalBlock
--- PASS: TestCreateDiagonalBlock (0.00s)
PASS
ok  	github.com/metrumresearchgroup/babylon/parsers/nmparser	0.013s
```

PR #13, opened 2019-08-01
Feature/table file

david-lyder:
This change supports a single TABLE statement that span lines.
Since there can be multiple TABLE statements, each with a FILE value, the OutputTable value should be a []string not a string. This work is in progress.

### DIFF
--- a/parsers/nmparser/structs.go
+++ b/parsers/nmparser/structs.go
@@ -16,7 +16,7 @@ type RunDetails struct {
 	NumberOfPatients    int64
 	NumberOfObs         int64
 	NumberOfDataRecords int64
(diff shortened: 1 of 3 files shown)


david-lyder:
added support for multiple outputTables

parsers/nmparser/structs.go
```diff
@@ -16,7 +16,7 @@ type RunDetails struct {
 	NumberOfPatients    int64
 	NumberOfObs         int64
 	NumberOfDataRecords int64
-	OutputTable         string
+	OutputTable         []string
```
dpastoor: should probably be renamed OutputTables (plural)

PR #14, opened 2019-08-02
feature/ofv

david-lyder:
…JECTIVE FUNCTION VALUE WITHOUT CONSTANT"

### DIFF
--- a/parsers/nmparser/parse_lst_file.go
+++ b/parsers/nmparser/parse_lst_file.go
@@ -5,27 +5,26 @@ import (
 	"strings"
 )
 
-func parseOFV(line string, ofvDetailsIn OfvDetails) OfvDetails {
-	ofvDetailsOut := ofvDetailsIn
+func parseOFV(line string, ofvDetails OfvDetails) OfvDetails {
 	if strings.Contains(line, "#OBJV:") {
 		result := strings.Replace(line, "*", "", -1)
-		ofvDetailsOut.OFVNoConstant, _ = strconv.ParseFloat(
+		ofvDetails.OFVNoConstant, _ = strconv.ParseFloat(
 			strings.TrimSpace(strings.Replace(result, "#OBJV:", "", -1)),
 			64)
 	} else if strings.Contains(line, "CONSTANT TO OBJECTIVE FUNCTION") {
-		ofvDetailsOut.OFV, _ = strconv.ParseFloat(
+		ofvDetails.OFV, _ = strconv.ParseFloat(
 			strings.TrimSpace(strings.Replace(line, "N*LOG(2PI) CONSTANT TO OBJECTIVE FUNCTION:", "", -1)),
 			64)
 	} else if strings.Contains(line, "OBJECTIVE FUNCTION VALUE WITHOUT CONSTANT") {
-		ofvDetailsOut.OFVNoConstant, _ = strconv.ParseFloat(
+		ofvDetails.OFVNoConstant, _ = strconv.ParseFloat(
 			strings.TrimSpace(strings.Replace(line, "OBJECTIVE FUNCTION VALUE WITHOUT CONSTANT:", "", -1)),
 			64)
 	} else if strings.Contains(line, "OBJECTIVE FUNCTION VALUE WITH CONSTANT") {
-		ofvDetailsOut.OFVWithConstant, _ = strconv.ParseFloat(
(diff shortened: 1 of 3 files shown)


parsers/nmparser/parse_lst_file.go
```diff
@@ -5,28 +5,31 @@ import (
...
-func parseOFV(line string) float64 {
-	// get rid of everything but stars and ofv
-	result := line[10:]
-	result = strings.Replace(result, "*", "", -1)
-	output, _ := strconv.ParseFloat(strings.Fields(result)[0], 64)
-	return output
+func parseOFV(line string, ofvDetails *OfvDetails) {
+	if strings.Contains(line, "#OBJV") {
+		// get rid of everything but stars and ofv
+		result := line[10:]
```
dpastoor: what's on the first 10 characters that makes it worth trimming off?

parsers/nmparser/parse_lst_file.go
```diff
@@ -5,28 +5,31 @@ import (
...
+			strings.TrimSpace(strings.Replace(line, "OBJECTIVE FUNCTION VALUE WITH CONSTANT:", "", -1)),
+			64)
+	}
 }
 
 // ParseLstEstimationFile parses the lst file
 func ParseLstEstimationFile(lines []string) LstData {
-
-	var ofvIndex int
+	ofvDetails := new(OfvDetails)
```
dpastoor: why `new` it vs just declaring the type `var ofvDeatils OfvDetails`

parsers/nmparser/parse_lst_file_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestParTestParseOBJV2(t *testing.T) {
+	var lines = []string{
+		"OBJECTIVE FUNCTION VALUE WITHOUT CONSTANT: 821.705",
+		"OBJECTIVE FUNCTION VALUE WITH CONSTANT:       1639.561",
+		"N*LOG(2PI) CONSTANT TO OBJECTIVE FUNCTION:    817.855",
+		"#OBJV:********************************************    -7913.528       **************************************************",
+	}
+	lstData := ParseLstEstimationFile(lines)
+	if lstData.OFV.OFV != 817.855 || lstData.OFV.OFVWithConstant != 1639.561 || lstData.OFV.OFVNoConstant != -7913.528 {
```
dpastoor: I'm not a fan of this `||` based testing structure. Namely, it does not scale well, especially if you have more complex structures to compare. In addition, it is unclear if you are testing the entire struct or just _parts_ of a struct. With testify you can just use `assert.Equal()` to compare against initialized structs with the correct value. 

```
expected := OfvDetails{
// correct values
}
assert.Equal(t, lstData.OFV, expected)
```

parsers/nmparser/parse_lst_file.go
dpastoor: aka probably worth noting as a comment something like,

line should look like #OBJV: *** <value> ***, though with many more stars. So they can see that you're trimming the #OBJV

david-lyder: This was existing code that I should have re-visited.  I've normalized this code so it follows the same pattern as the others.

parsers/nmparser/parse_lst_file.go
david-lyder: It's a pointer. The pattern is to call the "parseOFV" function repeatedly while looping through the lines, each time possibly updating one of the 3 values in the OfvDetails structure.

parsers/nmparser/parse_lst_file_test.go
david-lyder: Yeah, assert is way better.

parsers/nmparser/parse_lst_file.go
```diff
@@ -46,7 +48,13 @@ func ParseLstEstimationFile(lines []string) LstData {
 		case strings.Contains(line, "0DEFAULT SIGMA BOUNDARY"):
 			endParameterStucturesIndex = i
 		case strings.Contains(line, "#OBJV"):
-			ofvIndex = i
+			parseOFV(line, ofvDetails)
```
dpastoor: could we actually change this to not take a pointer receiver, and instead use a struct directly and just return a modified version, to keep with the functional programming style of most of the parsing logic we have.

something like:

```
ofvDetails = parseOFV(line, ofvDetails)
```

PR #16, opened 2019-08-07
refactor summary outputs

dpastoor:


david-lyder:
Adding this for review. Using json.Marshall\Unmarshall we can test.Assert on the populated structures. This could be useful after an initial review of the data for completeness. The location of this test file (summart_test.go) and the NONMEM files is TDB.

david-lyder:
I propose we create a new folder named "end-to-end-test" (or similar), move the file summary_test.go to that folder, and then use the git command "submodule" to link to a different repo for models, as discussed previously:

a repo called "example-models"
  data/
  stan/
  nonmem/
  mrgsolve/
The "example-models" repo is a curated asset, with well named files, etc.

submodule may not be the best solution for the implied dependency, but looks like it would work:
https://stackoverflow.com/questions/7813030/how-can-i-have-linked-dependencies-in-a-git-repo

testdata/example-models/nonmem/IOVMM/IOVMM.json
```diff
@@ -0,0 +1,155 @@
...
+					0,
+					0,
+					0.0359252
+				],
+				"sigma": [
+					0.0458045,
+					0
+				]
+			},
+			"std_err": {},
```
dpastoor: hmm we'll need to think about this object - do we want it to be an empty object vs an object with all zeros. Basically, we might want to initialize the stderr blocks to be all zeros of the same length as the parameter values to make sure the structure is always there

testdata/example-models/nonmem/IOVMM/IOVMM.json
```diff
@@ -0,0 +1,155 @@
...
+			"6. PROB POP 1"
+		]
+	},
+	"ofv": {
+		"ofv": 23157.251036757752,
+		"ofv_no_constant": 17485.751,
+		"ofv_with_constant": 40643.001566747465
+	},
+	"shrinkage_details": {
+		"eta": {},
```
dpastoor: hmm why didn't shrinkage get measured

dpastoor: I don't think shrinkage requires stderrs to be estimated but i may be wrong

testdata/summary_test.go
```diff
@@ -0,0 +1,98 @@
...
+		{
+			modFile:    "./example-models/nonmem/TMDD/1.mod",
+			goldenFile: "./example-models/nonmem/TMDD/TMDD.json",
+		},
+		{
+			modFile:    "./example-models/nonmem/NonLinearCL/1.mod",
+			goldenFile: "./example-models/nonmem/NonLinearCL/NonLinearCL.json",
+		},
+	}
+	bbiExe := "/usr/local/bin/bbi"
```
dpastoor: this should detect bbi on the path rather than hard coded.

testdata/summary_test.go
```diff
@@ -0,0 +1,98 @@
...
+		// unmarshal golden json
+		var goldenSummary parser.ModelOutput
+		err = json.Unmarshal(goldenJSON, &goldenSummary)
+		assert.Equal(t, nil, err, fmt.Sprintf("[%s] fail to unmarshal goldenSummary: %s", context, err))
+
+		// marshal golden objects to bytes
+		goldenParametersData, err := json.Marshal(goldenSummary.ParametersData)
+		assert.Equal(t, nil, err, fmt.Sprintf("[%s] fail to marshal golden ParametersData: %s", context, err))
+
+		goldenParametersStructures, err := json.Marshal(goldenSummary.ParameterStructures)
```
dpastoor: whats the purpose of doing all these remarshalls?

testdata/summary_test.go
```diff
@@ -0,0 +1,98 @@
...
+		stdout, err := exec.Command(bbiExe, "summary", "--tree", tt.modFile).Output()
+		assert.Equal(t, nil, err, fmt.Sprintf("[%s] fail exec %s: %s", context, bbiExe, err))
+
+		// unmarshal summary output
+		var bbiSummary parser.ModelOutput
+		err = json.Unmarshal(stdout, &bbiSummary)
+		assert.Equal(t, nil, err, fmt.Sprintf("[%s] fail to unmarshal summary: %s", context, err))
+
+		// marshal summary data to bytes and compare to golden data
+		summaryParametersData, err := json.Marshal(bbiSummary.ParametersData)
```
dpastoor: why marshal? - why not just let you directly assert bbiSummary to goldenSummary?

testdata/example-models/nonmem/IOVMM/IOVMM.json
david-lyder: There is no data for the shrinkage info in any of the output files. We currently expect strings like this: "ETASHRINKSD(%)" to populate the object.  

The only lines with "shrink" in them are these (from the lst and xml):
IN SHRINKAGE (ETASTYPE):             NO

testdata/summary_test.go
david-lyder: ok, but at some point I think we need to be clear about: (1) tools being installed correctly, and (2) executing the tool from the installed folder by default.  It's easier to support a tool that includes its own installer and users should only expect support when they have they have used the installer - which presumably installs the latest supported version to a known location. That's not to say the tool can only be installed in a certain location, but we can start with a recommendation.  BTW, in the krn\csh shell scripts I've written in the past, it was always risky to rely on the users PATH to get the correct version.

testdata/summary_test.go
david-lyder: I thought viewing differences would be easier with strings\bytes, but "go" actually does a very good job showing differences between objects. I'll remove the extra marshaling.

dpastoor:
LGTM!

PR #24, opened 2019-09-11
Refactor/summary

david-lyder:


david-lyder:
I think comparing the size of the arrays is sufficient to determine if the stderr array should be populated with 0's ... 
maybe a better test is:   if  len(estimate) > 0 && len(stderr) == 0

dpastoor:
I think either one would probably be fine, its definitely doing what I had in mind now

PR #25, opened 2019-09-12
feature/replace-termtable

david-lyder:
wip: replace termtable with tablewriter, can't seem to change color of individual rows


parsers/nmparser/summary.go
```diff
@@ -36,38 +35,38 @@ func (results ModelOutput) Summary() bool {
...
 		}
 
+		thetaTable.Append([]string{
+			string("TH " + strconv.Itoa(i+1)),
+			results.ParameterNames.Theta[i],
+			strconv.FormatFloat(numResult, 'f', -1, 64),
+			fmt.Sprintf("%s (%s%%)", strconv.FormatFloat(seResult, 'f', -1, 64), strconv.FormatFloat(rse, 'f', 1, 64))})
+	}
+
+	if anyRseOver {
```
dpastoor: we don't really want the entire column to be red, we'd just want the particular cells to be red.

That said, if some things aren't working (per the commit note), this is low priority and as long as we have tables period thats good enough for now!

david-lyder: tablewriter only supports coloring the entire column. 

In the first implementation, the entire column is colored if any of the row values exceed the criterion; in the second iteration(below), only the individual cell is colored. This could be extended to color the entire row.

david-lyder:
I am not seeing the latest changes of this file on release/v2.0.0:https://github.com/metrumresearchgroup/babylon/blob/e2827da22439bc78adb3da339730ffbed75be883/parsers/nmparser/summary.go#L22

PR #28, opened 2019-09-16
feature/replace-termtable

david-lyder:
all other changes to support tablewriter were merged, this change is only to use aurora for text color


dpastoor:
this was all merged into the v2.0.0 branch separately so can be closed.

Issue #22, opened 2019-08-26
give common heuristics


dpastoor:
# Summary
representative example via PsN:

```
Successful minimization [ OK ]
No rounding errors [ OK ]
No zero gradients [ OK ]
No final zero gradients [ OK ]
Hessian not reset [ OK ]
Parameter(s) near boundary [ WARNING ]
OMEGA(2,2) 0.000112 0
Covariance step [ OK ]
Large condition number [ WARNING ]
Correlations [ OK ]
```

To exercise summary:
from: github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/BQL
bbi summary --json 2

# Tests
- summary_test.go
     - TestSummary

dpastoor:
* [ ] Successful minimization [ OK ]

search for: 

```
0MINIMIZATION SUCCESSFUL
```

* [ ] No rounding errors [ OK ]
* [ ] No zero gradients [ OK ]
* [ ] No final zero gradients [ OK ]
* [ ] Hessian not reset [ OK ]
* [ ] Parameter(s) near boundary [ WARNING ]
* [ ] OMEGA(2,2) 0.000112 0
* [ ] Covariance step [ OK ]
* [ ] Large condition number [ WARNING ]
* [ ] Correlations [ OK ]

david-lyder:
### After change:

"run_heuristics": {
		"large_condition_number": "HeuristicFalse",
		"correlations_ok": "HeuristicTrue",
		"has_final_zero_gradient": "HeuristicFalse",
		"minimization_successful": "HeuristicTrue"
	},

david-lyder:
The above heuristic output is considered complete for version 2.0.0
Further enhancements will be addressed based on feedback from the scientists.

PR #29, opened 2019-09-16
feature/heuristics

david-lyder:
I think the best pattern to support this behavior is to define the structure members as Boolean  pointers. With pointers, the JSON tag "omitempty" will return true, false or empty (when nil). If they are just bools (not pointers), then empty is returned when the value is false.

issue #22


parsers/nmparser/structs.go
```diff
@@ -31,7 +31,14 @@ type ParametersData struct {
 	Fixed            ParametersResult   `json:"fixed,omitempty"`
 }
 
+// RunHeuristics ...
+// some values are defined as pointers to support tri-state: true, false, nil
 type RunHeuristics struct {
+	CovarianceStepOmitted *bool `json:"covariance_step_omitted,omitempty"`
```
dpastoor: why pointers? there doesn't seem to be any reason for pointers and breaks the consistency of the rest of the struct api

parsers/nmparser/parse_lst_file.go
```diff
@@ -76,6 +125,11 @@ func parseGradient(lines []string) (hasZero *bool, hasFinalZero *bool) {
 	return anyZero, anyFinalZero
 }
 
+func newBool(value bool) *bool {
```
dpastoor: this feels like an antipattern around the use of pointers

parsers/nmparser/parse_lst_file.go
```diff
@@ -54,6 +55,54 @@ func parseOFV(line string, ofvDetails OfvDetails) OfvDetails {
...
+			eigenvalues = append(eigenvalues, eigenvalue)
+		}
+	}
+
+	if len(eigenvalues) >= 2 {
+		sort.Float64s(eigenvalues)
+		ratio := eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
+		// TODO: get largeConditionNumber threshold from config
+		// or derive. something like (number of parameters) * 10
+		if ratio > 1000.0 {
```
dpastoor: we should probably expose this better, so it can be controlled via configuration, so for now would elevate it to a param

dpastoor:
I'm just seeing the top level comment around the pointer choice

> I think the best pattern to support this behavior is to define the structure members as Boolean pointers.

> just bools (not pointers), then empty is returned when the value is false.

why not just remove omitempty. That is just a default for the code gen to add in json? that should return the false the same, but not go through these pointer hoops

david-lyder:
To satisfy omitting JSON values that were not found, we need 3 choices for a bool value so we can distinguish between false-never-set and false-explicitly-set. 

The difference is between finding data and parsing to determine a false value vs. not finding the data at all.

For example, if we don't use "omitempty", we'll get false for HasZeroGradient, HasFinalZeroGradient, and LargeConditionNumber even if we do not find the data to parse.  

If we use "omitdata", none of the those values (HasZeroGradient, HasFinalZeroGradient, and LargeConditionNumber) will be included in the JSON even if the data is found and parsed and determined to be false.  

For correlationOK, with "omitdata" data not set, we would return false even if the data is not found.

Using a bool type, we'll either return false for data we never checked, or omit values for data we have checked. With a bool pointer, we can make a determination to include on a case-by-case basis.
Let me know what you prefer.

parsers/nmparser/structs.go
```diff
@@ -31,9 +31,16 @@ type ParametersData struct {
 	Fixed            ParametersResult   `json:"fixed,omitempty"`
 }
 
+// RunHeuristics ...
+// some values are defined as pointers to support tri-state: true, false, nil
 type RunHeuristics struct {
-	HasZeroGradient        bool `json:"has_zero_gradient,omitempty"`
-	MinimizationSuccessful bool `json:"minimization_successful,omitempty"`
+	CovarianceStepOmitted  string `json:"covariance_step_omitted,omitempty"`
```
dpastoor: these should not be strings. They should be typed properly. Please refer to implementations such as:

https://gist.github.com/lummie/7f5c237a17853c031a57277371528e87

and 

https://github.com/campoy/jsonenums

parsers/nmparser/get_model_output.go
```diff
@@ -12,7 +12,7 @@ import (
 // GetModelOutput populates and returns a ModelOutput object by parsing files
 // ParameterData is parsed from the ext file when useExtFile is true
 // ParameterData is parsed from the lst file when useExtFile is false
-func GetModelOutput(filePath string, verbose bool, useExtFile bool) ModelOutput {
+func GetModelOutput(filePath string, verbose bool) ModelOutput {
```
dpastoor: why did you remove the ability to use the ext file or not. What if the user knows explicitly they don't have it or wants to suppress using it. Attempting to read and blowing up is not a suitable solution

parsers/nmparser/structs.go
david-lyder: That is pretty much the implementation that I have, but when the JSON renders, the values are ints. i

parsers/nmparser/get_model_output.go
david-lyder: I thought you were onboard with "if the files exists, use it".  BTW, the flag was always true in the last implementation.

dpastoor: yes, as a default! not an only way. there is no reason for us not to expose turning it off if needed. What if its there but can't be parsed correctly, for example with #31 what if they want to compare the summary from the lst vs the ext file as they suspect a bug but aren't sure, etc.

david-lyder: Ok, so do we want a flag for grd as well?

parsers/nmparser/structs.go
dpastoor: where is your marshalling and unmarshalling functionality? 

> but when the JSON renders, the values are ints.

where?

parsers/nmparser/get_model_output.go
dpastoor: yep - and any other file we pull in that we could also get those values from the lst file, if we need to.

I see the lst file only parsing as a fallback that we will need to monitor in the future based on user need. My hope is that we can actually move the parsing out to all these much more consistent files, and parse as little as possible from the lst.

parsers/nmparser/structs.go
david-lyder: Not sure I understand the question, it's serialized using the JSON annotations:
CovarianceStepOmitted  string `json:"covariance_step_omitted,omitempty"`

If the type is an enum/iota, it's serialized as an int

dpastoor: its only serialized serialized as an int _if you haven't defined an unmarshal method_. Thats why "That is pretty much the implementation that I have" makes no sense. You literally haven't implemented the key functionality for serializing and deserializing. 

Instead you're coercing to a string that has no type stability on it and would mean that we would have to put conditional logic anywhere else that uses those structs to string match rather than refer to the enum directly.

instead of type stable checks like

```
if RunHeuristics.Minimization == StatusTrue {
...
}
```

you'd need

```
RunHeuristics.Minimization == "True"
```

as you're ToString()'ing everything, which defeats the entire purpose.

It should stay as an enum type until serialization, and upon deserialization should properly reflect the type as well. Thus you need Unmarshal/Marshal methods.

parsers/nmparser/parse_lst_file.go
```diff
@@ -222,28 +293,28 @@ func ParseLstEstimationFile(lines []string) ModelOutput {
...
-			runHeuristics.CovarianceStepOmitted = True.String()
+			runHeuristics.CovarianceStepOmitted = HeuristicTrue
 		case strings.Contains(line, "EIGENVALUES OF COR MATRIX OF ESTIMATE"):
 			// TODO: get largeNumberLimit from config
 			// or derive. something like (number of parameters) * 10
-			runHeuristics.LargeConditionNumber = setLargeConditionNumber(lines, i, 1000.0).String()
+			runHeuristics.LargeConditionNumber = setLargeConditionNumber(lines, i, 1000.0)
 		case strings.Contains(line, "CORRELATION MATRIX OF ESTIMATE"):
-			runHeuristics.CorrelationsOk = setCorrelationsOk(lines, i).String()
+			runHeuristics.CorrelationsOk = setCorrelationsOk(lines, i, 0.95)
```
dpastoor: where'd you get the 0.95 from?

parsers/nmparser/structs.go
```diff
@@ -127,26 +129,38 @@ type ExtData struct {
...
-// 	}
-// 	return false
-// }
+var toID = map[string]HeuristicStatus{
+	"HeuristicUndefined": HeuristicUndefined,
+	"HeuristicTrue":      HeuristicTrue,
+	"HeuristicFalse":     HeuristicFalse,
+}
+
+// MarshalJSON marshals the enum as a quoted json string
```
dpastoor: we will need to add an unmarshal json as well I believe, especially for our tests that will use golden json files as the source of truth

parsers/nmparser/parse_lst_file.go
```diff
@@ -107,11 +291,31 @@ func ParseLstEstimationFile(lines []string) ModelOutput {
...
+		case strings.Contains(line, "RESET HESSIAN"):
+			runHeuristics.HessianReset = HeuristicTrue
+		case strings.Contains(line, "PARAMETER ESTIMATE IS NEAR ITS BOUNDARY"):
+			runHeuristics.ParameterNearBoundary = HeuristicTrue
+		case strings.Contains(line, "COVARIANCE STEP OMITTED: NO"):
+			runHeuristics.CovarianceStepOmitted = HeuristicTrue
+		case strings.Contains(line, "EIGENVALUES OF COR MATRIX OF ESTIMATE"):
+			// TODO: get largeNumberLimit from config
+			// or derive. something like (number of parameters) * 10
+			runHeuristics.LargeConditionNumber = getLargeConditionNumberStatus(lines, i, 1000.0)
```
dpastoor: how did you choose a value of 1000?

parsers/nmparser/parse_lst_file.go
```diff
@@ -107,11 +291,31 @@ func ParseLstEstimationFile(lines []string) ModelOutput {
...
+		case strings.Contains(line, "CORRELATION MATRIX OF ESTIMATE"):
+			// TODO: get correlationLimit from config
+			runHeuristics.CorrelationsOk = getCorrelationStatus(lines, i, 0.95)
```
dpastoor: how did you choose a value of 0.95?

david-lyder: PsN default it 0.9, so I'll change to that.

https://github.com/UUPharmacometrics/PsN/blob/88320646310b63f0c88075a6a71f4d5a87be4528/bin/sumo#L59

Also, from sumo doc:
-correlation_limit = N
Default is 0.9. The cut-off for issuing a warning for high correlations between the parameter estimates. This test will only
be made if the $COV step has been requested in the NMtran
control stream.
The default is set in psn.conf.

parsers/nmparser/parse_lst_file.go
david-lyder: PsN default is 1000.

https://github.com/UUPharmacometrics/PsN/blob/88320646310b63f0c88075a6a71f4d5a87be4528/bin/sumo#L60

From sumo doc:
-condition_number_limit = N
Default is 1000. The cut-off for issuing a warning for a high condition number. The test will only be made if the $COV step has
been requested in the NMtran control stream and NONMEM
has been asked to report eigenvalues ($COV PRINT=E).

PR #33, opened 2019-09-23
Issue/index out of range

david-lyder:
Added protection around zero length shrinkage array.


dpastoor:
new changes look good - are the tests updated to reflect this change? and do we have coverage for both present/not present to autotest from now on?

david-lyder:
The feature/heuristics branch has end-to-end tests at testdata/summary_test.go, I think that would be the most appropriate location for this testing. The file does not exist in this branch (issue/index-out-of-range)
For now, I'll develop the test in a new file. Later we can do some git magic, if needed, to merge it all together.

david-lyder:
Updated the golden JSON file to include zero-filled shrinkage_details records. When the test is executed from summary_test.go it should pass. See issue #33 for screenshots.

Issue #52, opened 2019-10-16
OMEGA parameters not parsed correctly


david-lyder:
### Summary
Not all mod/lst files have lines defining the size and structure of OMEGA. The current code looks for and parses the lines:
0OMEGA HAS SIMPLE DIAGONAL FORM WITH DIMENSION
or
0OMEGA HAS BLOCK FORM:
The result of not finding the "0OMEGA HAS " causes an index OOB error inparseParameterStructure as an array of zero lines is passed into the function.

**Proposal**
The length of OMEGA can be captured from the ext file (for example as OmegaCount). Then the first # of OmageCount lines in the lst file that start with "$OMEGA BLOCK" can be captured for the OMEGA values and possible the parameter name.  Note, these 3 (known) forms of "OMEGA BLOCK" must be supported:
**(1)**
$OMEGA BLOCK(3)
1                ;1 - CL 
2   3          ;2 - V1 
4   5   6    ;3 - V2 
**(2)**
OMEGA BLOCK(1) 0 FIX   ;4 - VMAX
**(3)**
$OMEGA BLOCK(1) SAME ;8 - IOVCL

### Tests
TestSummary
TestSummaryTable
note: after adding support for a new control stream

david-lyder:
### Before change:

bbi summary 117.mod
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/metrumresearchgroup/babylon/parsers/nmparser.parseParameterStructure(0xc0001a98b0, 0x0, 0x375, 0xe, 0xc000162970, 0x1)
	/Users/davidl/go/src/github.com/metrumresearchgroup/babylon/parsers/nmparser/parse_parameter_structure.go:20 +0x223
github.com/metrumresearchgroup/babylon/parsers/nmparser.ParseParameterStructures(0xc0001a98b0, 0x4, 0x375, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/davidl/go/src/github.com/metrumresearchgroup/babylon/parsers/nmparser/parse_parameter_structure.go:54 +0x35d
github.com/metrumresearchgroup/babylon/parsers/nmparser.ParseLstEstimationFile(0xc0001a8000, 0x402, 0x500, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[...]

PR #53, opened 2019-10-17
feature/omega from ext

david-lyder:
changes to support #52 
The reason we don't **_always_** use the ext file to get Omega info is because the lst file has more information than the ext file. The ext file only provides the length (dimension) of the Omega block.

### DIFF
--- a/parsers/nmparser/get_model_output.go
+++ b/parsers/nmparser/get_model_output.go
@@ -3,6 +3,7 @@ package parser
 import (
 	"log"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/metrumresearchgroup/babylon/utils"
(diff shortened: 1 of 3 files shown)


parsers/nmparser/get_model_output.go
```diff
@@ -33,8 +34,14 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
 			panic(err)
 		}
-		extData, _ := ParseExtData(ParseExtLines(extLines))
+		extData, parameterNames := ParseExtData(ParseExtLines(extLines))
 		results.ParametersData = extData
+
+		// check for Omega
+		if results.ParameterStructures.Omega == nil {
+			dim := getDimension(parameterNames.Omega[len(parameterNames.Omega)-1])
+			results.ParameterStructures.Omega = createDiagonalBlock(dim)
```
dpastoor: this will not be the same parameterstrucutres type if it parses from the lst file as its doing a simple createDiagonalBlock

https://github.com/metrumresearchgroup/babylon/blob/test/nonmem/parsers/nmparser/parse_block_structure.go#L9:6

Look at the test results https://github.com/metrumresearchgroup/babylon/blob/test/nonmem/parsers/nmparser/parse_block_structure_test.go 

This currently would not do that

david-lyder: The est file is only used when there is no data found in the lst file.  If we have block data we use it, if not we can create the diagonal based on the number of Omegas in the est.

PR #55, opened 2019-10-17
shrinkage and testing refactor

david-lyder:
please review proposed structure


parsers/nmparser/read_shk_test.go
```diff
@@ -0,0 +1,66 @@
...
+		assert.Equal(t, tt.expected1, ed.EstimationMethods[0], "Fail :"+tt.context)
+		assert.Equal(t, tt.expected2, ed.EstimationMethods[1], "Fail :"+tt.context)
+		assert.Equal(t, "TYPE", ed.ParameterNames[0], "Fail :"+tt.context)
+		assert.Equal(t, "ETA(4)", ed.ParameterNames[5], "Fail :"+tt.context)
+		assert.Equal(t, "1            1 -3.67026E-03 -3.91460E-03 -8.00654E-03 -1.95727E-03", ed.EstimationLines[0][0], "Fail :"+tt.context)
+
+		shk := ParseShrinkage(ed.EstimationLines[0])
+		assert.Equal(t, -0.00367026, shk.EtaBar[0], "Fail :"+tt.context)
+
+		sd := ParseShkData(ed)
```
dpastoor: should also include a dimension check (eg that only 4 values per slice are captured in this test)

You should also change the values in table 2 to make sure you're not accidentally just reusing the same table multiple times - eg how can you tell if you a bug isn't introduced to make the tables out of order in the array or something.

parsers/nmparser/read_shk.go
```diff
@@ -0,0 +1,113 @@
...
+			case "9":
+				for n := 2; n < (length); n++ {
+					shrinkageDetails.EpsVR = append(shrinkageDetails.EpsVR, strToFloat(fields[n]))
+				}
+			case "10":
+				for n := 2; n < (length); n++ {
+					shrinkageDetails.EbvVR = append(shrinkageDetails.EbvVR, strToFloat(fields[n]))
+				}
+			default:
+				// TODO: logrus
```
dpastoor: should issue a parsing warning but continue

parsers/nmparser/read_shk_test.go
david-lyder: ok. added a table 3 with different values in the first row.

dpastoor:
After short review, this PR currently has major issues:

- tests were re-run under nonmem 7.4
- shk file do not represent the entire new shk structure given a model without covariance step run (eg pval,etc should also have slice of 0's)

This will require a much more careful review and is not mergeable in its current state

parsers/nmparser/get_model_output.go
```diff
@@ -95,6 +95,12 @@ func GetModelOutput(filePath string, verbose, noExt, noGrd, noCov, noCor, noShk
 }
 
 func setDefaultValues(results *ModelOutput, etaCount, epsCount int) {
+	// method name
+	for i := range results.ParametersData {
+		if len(results.ParametersData[i].Method) == 0 {
+			results.ParametersData[i].Method = fmt.Sprintf("Method%v", i+1)
```
dpastoor: this might be incorrect accidentally - if this came back as Method1 due to a default, Method1 is "FOCE". IMO the default method should be "METHOD NOT DETECTED" or something along those lines

parsers/nmparser/parse_run_details.go
```diff
@@ -42,21 +42,21 @@ func parseLine(line string, n int) string {
...
-	version := ""
-	runStart := ""
-	runEnd := ""
-	estimationTime := 0.0
-	covarianceTime := 0.0
-	functionEvaluations := int64(0)
-	significantDigits := 0.0
-	problemText := ""
-	modFile := "" // TODO, pass in from caller
+	version := DefaultString
```
dpastoor: per style guide: https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations

this should be grouped and use a var( declaration)

```
var (
 runStart = DefaultString
 runEnd = DefaultString
...
) 
```

and should be organized by type

dpastoor: at this point, it might make more sense anyway just to initialize RunDetails with NewRunDetails() then set things on that directly and completely get rid of all these variables

david-lyder: I agree, I was very close to doing that ...

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
 		}
 
-		if len(results.ShrinkageDetails[n].Eps.SD) == 0 {
-			dim := 0
-			for i := range results.ParameterStructures.Sigma {
-				if results.ParameterStructures.Sigma[i] > 0 {
-					dim++
+		// shrinkage eps's
+		if epsCount > 0 {
+			for n := 0; n < len(results.ShrinkageDetails); n++ {
```
dpastoor: range

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
-	for n := 0; n < len(results.ShrinkageDetails); n++ {
-		if len(results.ShrinkageDetails[n].Eta.SD) == 0 {
-			dim := 0
-			for i := range results.ParameterStructures.Omega {
-				if results.ParameterStructures.Omega[i] > 0 {
-					dim++
-				}
+	// shrinkage eta's
+	if etaCount > 0 {
+		for n := 0; n < len(results.ShrinkageDetails); n++ {
```
dpastoor: range

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
+		var nameCount int
+
+		thetaCount := len(results.ParametersData[i].Estimates.Theta)
+		omegaCount := len(results.ParametersData[i].Estimates.Omega)
+		sigmaCount := len(results.ParametersData[i].Estimates.Sigma)
+
+		// stderr
+		nameCount = len(results.ParametersData[i].StdErr.Theta)
+		for n := nameCount; n < thetaCount; n++ {
+			results.ParametersData[i].StdErr.Theta = append(results.ParametersData[i].StdErr.Theta, 0)
```
dpastoor: I'm not sure what this is doing?

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
 
+	epsCount := 0
+	for i := range results.ParameterStructures.Sigma {
+		if results.ParameterStructures.Sigma[i] > 0 {
+			epsCount++
+		}
+	}
+	etaCount := 0
+	for i := range results.ParameterStructures.Omega {
+		if results.ParameterStructures.Omega[i] > 0 {
```
dpastoor: the parameter structure should be nonzero for non-diagonal elements of omega though if the user uses an $OMEGA BLOCK(X) form though? I don't think this logic is correct. The etas should correspond to the diagonal elements of the omega matrix

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
 
-	for n := 0; n < len(results.ShrinkageDetails); n++ {
-		if len(results.ShrinkageDetails[n].Eta.SD) == 0 {
-			dim := 0
-			for i := range results.ParameterStructures.Omega {
-				if results.ParameterStructures.Omega[i] > 0 {
-					dim++
-				}
+	// shrinkage eta's
+	if etaCount > 0 {
```
dpastoor: it looks like this should use a constructor function rather than all these sub loops. TOo much duplication

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
+	if !noShk {
+		shkFilePath := strings.Join([]string{filepath.Join(dir, runNum), ".shk"}, "")
+		shkLines, err := utils.ReadLines(shkFilePath)
+		if err != nil {
+			panic(err)
+		}
+		results.ShrinkageDetails = ParseShkData(ParseShkLines(shkLines), etaCount, epsCount)
+	}
+
+	setDefaultValues(&results, etaCount, epsCount)
```
dpastoor: Can this be more descriptive - setMissingValuesToDefault() or something along those lines.

parsers/nmparser/get_model_output.go
dpastoor: this definitely needs some inline comments explaining the top level logic of whats going on here

parsers/nmparser/get_model_output.go
```diff
@@ -67,48 +68,137 @@ func GetModelOutput(filePath string, verbose bool, noExt bool, noGrd bool, noCov
...
+	return results
+}
+
+func setDefaultValues(results *ModelOutput, etaCount, epsCount int) {
```
dpastoor: this definitely needs tests to make sure the logic is behaving correctly

parsers/nmparser/parse_run_details.go
```diff
@@ -101,27 +87,11 @@ func ParseRunDetails(lines []string) RunDetails {
 		}
 	}
 
-	if version == "7.4.3" {
-		if runStart == "" {
-			runStart = lines[0]
+	if runDetails.Version == "7.4.3" {
```
dpastoor: too specific, should just be if nonmem 7.4 vs 7.3. What will happen when nonmem 7.4.4 comes out

parsers/nmparser/get_model_output.go
david-lyder: Go uses a copy of the value instead of the value itself within a range clause, so the array is lost when exiting the scope of the range.

parsers/nmparser/get_model_output.go
david-lyder: Go uses a copy of the value instead of the value itself within a range clause, so the array is lost when exiting the scope of the range.

parsers/nmparser/get_model_output.go
david-lyder: OK, I'll add comments to this effect:
if the theta "stderr" array has less values than the theta "estimates" array, pad with 0's.

wip to use default values instead of 0's.  BTW,  there are two kinds of defaults values:
1. slices, which can't be done until dimensions are known (somehow)
2. values, it's better to initialize these when the object is created and then allow the default to be overwritten

parsers/nmparser/get_model_output.go
david-lyder: yes, a unit test is required

parsers/nmparser/parse_run_details.go
david-lyder: ok (that LOC is pretty old)

parsers/nmparser/get_model_output.go
david-lyder: a constructor for the float, like NewFloat? 
The dimension of the arrays are generally not known when the object is created. 
This code needs to be updated to change the 0's to a default value's, as part of that effort, I can investigate if the dimensions could be captured before instantiation

parsers/nmparser/get_model_output.go
david-lyder: done

parsers/nmparser/get_model_output.go
david-lyder: comments have been added

parsers/nmparser/get_model_output.go
david-lyder: We are looking for the number of etas here to possibly pad ParameterStructures.Omega[].
If this is a valid use case, we can't useParameterStructures.Omega[] to get the number of etas.

parsers/nmparser/get_model_output.go
dpastoor: We are using range all over the place in this very function....

for n := range results.ShrinkageDetails ....

You are correct if. we were ranging over the values themselves but we're not, we're accessing by index. 

Can you explain why you see this different than like 106

for i :=range results.ParameterData ...

such that you can't use range?

david-lyder: I think misunderstood your first comment  - thought you were asking to range over the ShrinkageDetail values. I'm sure we can range over the indices like on line 106.

david-lyder:
I still need to do a final review of the golden files. Viewing the file differences alone is not sufficient as there are many. I need to compare the values to their source (lst, ext, shk ..)

parsers/nmparser/get_model_output.go
david-lyder: I don't have an exhaustive list of all of the use-cases in which NONMEM can fail. This is a general solution since we know the dimensions _should_ be equal. Here is an example from the TMDD lst file:


THETA - VECTOR OF FIXED EFFECTS PARAMETERS   *********


         TH 1      TH 2      TH 3      TH 4      TH 5      TH 6      TH 7      TH 8      TH 9     
 
         2.68E-01  2.17E+00  1.44E-01  1.19E+00  5.57E+01  1.96E+00  8.00E-02  1.50E-01  1.05E+01


THETA - VECTOR OF FIXED EFFECTS PARAMETERS   *********


         TH 1      TH 2      TH 3      TH 4      TH 5      TH 6      TH 7      TH 8      TH 9     
 
         1.57E-02  1.61E-01  5.51E-03  3.26E-02  3.71E+00  3.97E-02  1.65E-03 ......... .........

parsers/nmparser/parse_lst_file.go
```diff
@@ -374,7 +387,7 @@ func ParseLstEstimationFile(lines []string) ModelOutput {
 		case strings.Contains(line, "0MINIMIZATION SUCCESSFUL"):
 			runHeuristics.MinimizationSuccessful = HeuristicTrue
 		case strings.Contains(line, "GRADIENT:"):
```
dpastoor: can you please add a piece of doc as to why this is needed. I believe this is because it looks like for larger number of params it can run onto a new line and thus you need to chase that down, but these "deviations" from the simple matching should get noted

david-lyder: Initially, this is the style I used (on purpose) for string matching, thinking I did not want to "fail to match" because of whitespace or other irrelevant characters, especially WRT to different control streams and different versions of NONMEM.  I've now seen that the lst text is pretty consistent, so an "equal-to" test would probably be fine.

dpastoor:
There are a number of issues present here that must be unrolled carefully, and this PR has become too unwieldy to unwind them. Going to re-review all code holistically

PR #83, opened 2019-11-03
Triage/bugfixes

dpastoor:


dpastoor:
multiple fixes, but still need re-eval of testing in broader perspective, but codebase better to start from

PR #89, opened 2019-11-25
Located the section adding the estimated string to the next job run

shairozan:
Found where the est value was being appended in `FindNextEstDirNum` and resolved that. Need to figure out what tests are valid and need to be adjusted.


shairozan:
Updated the tests that were failing for Selecting the name of the new directory

dpastoor:
consolidation woop woop

PR #91, opened 2019-12-04
Beginning turnstile implementation and getting basics working. Need t…

shairozan:
In the process of working to grid submission, implementing turnstile such that each target (local, sge, etc) will follow the same workflow.


cmd/nonmem.go
```diff
@@ -0,0 +1,395 @@
...
+		"worker.set",
+		"xmloff.set",
+		"fort.2001",
+		"fort.2002",
+		"flushtime.set",
+	}
+
+	//Get the files from the model
+	//Look specifically for .mod file
+	fileContents, err := ioutil.ReadFile(path.Join(filepath, filename+".mod"))
```
dpastoor: we need to actually expand any places that we look for `.mod` to also look for `.ctl` --> .mod = psn convention, .ctl is general nonmem convention metrum uses

cmd/root_darwin.go
```diff
@@ -0,0 +1,3 @@
+package cmd
+
+const lineEnding string = "\n"
```
dpastoor: these are all + more already captured in os

https://golang.org/pkg/os/#pkg-constants

os.PathSeparator

runner/clean_and_copy_est.go
```diff
@@ -44,10 +68,12 @@ func CleanEstFolderAndCopyToParent(
 ) error {
 	outputFiles := EstOutputFileCleanLevels(runNum)
 	keyOutputFiles := EstOutputFilesByRun(runNum)
-	var copiedFiles []CopiedFile
+	var copiedFiles []TargetedFile
 	if !filepath.IsAbs(dirToClean) {
 		dirToClean = filepath.Join(parentDir, dirToClean)
 	}
+
+	//We're literally forcably incrementing the cleanlvl / keep level to make sure they're not operated on?
```
dpastoor: yep - that was my "trick", rather than doing an incremental build up of keep files through various logic, instead just mutate that global list to adjust it so it gives the desired behavior - not wed to it, just some background to how I did it the first time.

cmd/root_darwin.go
shairozan: I'm not looking for a path separator, I'm looking for the line endings specific to the OS, specifically when writing the gitignore file. This is a workaround I'd seen some people use since I couldn't find an os.LineEnding like I'm used to in C# or Java

cmd/nonmem.go
shairozan: That makes sense. I can just iterate through "expected" extensions to find any files

cmd/nonmem.go
```diff
@@ -0,0 +1,395 @@
...
+	fileLines := strings.Split(string(fileContents), "\n")
```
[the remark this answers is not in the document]
shairozan: I'll have to revisit this then. I originally was using os.Copy, but the contents of all the files would be empty, thus I'd move to read / write. I'll have to dig into platform related issues for reading / writing files.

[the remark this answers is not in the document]
dpastoor: look at the goutils - its basically just replacing the ReadFile + split with better logic. not much more code on our end, just using better logic

From the scanner docs:

> ScanLines is a split function for a Scanner that returns each line of text, stripped of any trailing end-of-line marker. The returned line may be empty. The end-of-line marker is one optional carriage return followed by one mandatory newline. In regular expression notation, it is \r?\n. The last non-empty line of input will be returned even if it has no newline.

cmd/root_darwin.go
dpastoor: whew I definitely had path separators + using scanner instead of splitting on a separator, and those things got mashed together the mind yesh.

dpastoor: follow up is why you need the different line endings? If we're not using that as a const to split on (since we should scan instead anyway) - the other option would be for writing line by line, in which i'm not sure we need to go that far - it seems like my understanding is just using \n is fine. It would also give the added benefit of us not having to deal with cross platform messiness with testing if we wanted to do just like byte-comparisons/hashes of expected vs actual output files since our output files would then only be `\n` separated.

runner/clean_and_copy_est.go
shairozan: It was more of a note for myself as I was walking through the movement of pieces. There were a lot of sections I wasn't totally certain about, so I had noted the details down to facilitate me coming back to them later. We actually need to sit down at some point and make sure the levels I have now make sense.

cmd/root_darwin.go
shairozan: It was a scenario where I was going to write bytes to a file as "*" for the gitignore file and realized that "Holy crap, I need to have a different line ending for that component" since windows uses the \r\n sequence as opposed to just \n. Originally I was just going to leave it at "\n" but it felt prudent to deal with it there since I had thought of it as a problem. It leads to a crap ton of other questions to because technically on windows gitignore is handled by their compact bash shell that still technically parses the "\n" so it shouldn't matter in the end.

cmd/root.go
```diff
@@ -66,11 +67,21 @@ func Execute(build string) {
 func init() {
 	cobra.OnInitialize(initConfig)
 
+	viper.SetConfigName(".babylon")
+	viper.SetConfigType("yaml")
+	// viper.AddConfigPath(".")
+
```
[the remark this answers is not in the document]
shairozan: It's actually being done that way kind of already. The defaults on init add configuration paths for $HOME and the current directory. In the case of absolute call to the model, however, once we have established the original path for the model, we check to see if a config has been loaded, if not we add the original path of the model in as a location and try to read it in again. That way you could make sure the config is committed with the models and it would work whether you call it locally, or absolutely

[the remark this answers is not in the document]
shairozan: I'll double check the spec though. I need to find a way to enforce priority for viper's configuration path searching

cmd/nonmem.go
shairozan: So what makes even more sense is to actually just pass down the full file (name+ext) to this method and extract the filename from it. This way we don't have to curse through a list of "known" extensions. It's just whatever the user specified in the args.

cmd/nonmem.go
```diff
@@ -0,0 +1,395 @@
...
+	return buf.Bytes(), nil
+}
+
+func buildNonMemCommandString(l NonMemModel) string {
+
+	// TODO: Implement cache
+	noBuild := false
+	nmExecutable := l.Settings.NmExecutableOrPath
+	cmdArgs := []string{
+		l.OutputDir + "/" + l.Model,
```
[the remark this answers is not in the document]
shairozan: Yeah dur. Was just trying to fit the pieces earlier on but should have cleaned it up afterwards

cmd/nonmem.go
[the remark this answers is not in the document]
shairozan: So the read write were already implemented in utils as you had mentioned. Just moved to using those to avoid the conflict.

cmd/local.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+//Monitor is unimplemented here. It's the 3rd phase of Turnstile execution
+func (l LocalModel) Monitor(channels *turnstile.ChannelMap) {
+	//Do nothing for this implementation
+}
+
+//Cleanup is the last phase of execution, in which computation / hard work is done and we're cleaning up leftover files, copying results around et all.
+func (l LocalModel) Cleanup(channels *turnstile.ChannelMap) {
+	time.Sleep(10 * time.Millisecond)
```
dpastoor: why sleep 10 ms?

cmd/local.go
```diff
@@ -0,0 +1,367 @@
...
+		}
+
+		err = utils.WriteLines(source, path.Join(pwi.FilesToCopy.CopyTo, file))
+
+		if err != nil {
+			log.Printf("An erorr occurred while attempting to copy the files: File is %s", v.File)
+			continue
+		}
+
+		v.File = l.Nonmem.OutputDir + "/" + v.File
```
dpastoor: filepath.Join?

cmd/local.go
```diff
@@ -0,0 +1,367 @@
...
+
+		copied = append(copied, v)
+	}
+
+	//Write to File in original path indicating what all was copied
+	copiedJSON, _ := json.MarshalIndent(copied, "", "    ")
+
+	afero.WriteFile(fs, path.Join(l.Nonmem.OriginalPath, "copied.json"), copiedJSON, 0750)
```
dpastoor: Thinking about this, we should only write this out if files are actually copied I think. This will keep from cluttering the root dir with a bunch of essentially empty copies.

Also, more importantly this looks to be incorrect. If you have a file named copied.json, multiple models will keep overwriting that file. It should be `{{.Name}}_copied.json`

cmd/local.go
```diff
@@ -0,0 +1,367 @@
...
+	lo.Models = localModelsFromArguments(args)
+
+	if len(lo.Models) == 0 {
+		log.Fatal("No models were located or loaded. Please verify the arguments provided and try again")
+	}
+
+	//Display Summary
+
+	//Models Added
+	log.Printf("A total of %d models have been located for work", len(lo.Models))
```
dpastoor:
```diff
- "A total of %d models have been located for work"
+ "A total of %d models have been located for execution"
```

cmd/local.go
```diff
@@ -0,0 +1,367 @@
...
+			}
+		}(m)
+	}
+
+	//Basically wait
+	for !m.IsComplete() {
+		time.Sleep(5 * time.Millisecond)
+	}
+
+	if len(lo.Models) > 0 {
```
dpastoor: above you're already checking if len(lo.Models) == 0 fatally log, so probably don't need to check again

cmd/local.go
```diff
@@ -0,0 +1,367 @@
...
+		if viper.GetBool("saveConfig") {
+			configlib.SaveConfig(lo.Models[0].Nonmem.OriginalPath)
```
dpastoor: what if models are coming from different directories? eg I submit two directories of models to run

cmd/sge.go
```diff
@@ -0,0 +1,312 @@
...
+		if !strings.Contains(string(output), "job is not allowed to run in any queue") {
+			//If the error doesn't appear to be the above error, we'll generate the concurrent error and move along
+			return newConcurrentError(model.Model, "Running the programmatic shell script caused an error", err)
+		}
+	}
+
+	err = afero.WriteFile(fs, path.Join(model.OutputDir, model.Model+".out"), output, 0750)
+
+	if err != nil {
+		return newConcurrentError(model.Model, "Having issues writing hte output file from command execution", err)
```
dpastoor:
```diff
- "Having issues writing hte output file from command execution"
+ "Having issues writing the output file from command execution"
```

cmd/local.go
shairozan: Was having issues with IO not actually happening consistently. IE when you 30-40 models running, what I would notice is that, without some kind of air gapping, some directories would cleanup and others wouldn't. More work was done afterwards to deal with it, so I can remove this and see if I still have issues.

cmd/local.go
shairozan: Both make total sense. We already have the model filename as part of the NonMemModel so we can just concatenate it to the filename. We'll also keep track of copied files, and if any are copied, create the file.

cmd/local.go
shairozan: Yeah this is kind of a holdover I need to cleanup. It was there after completion during initial design to make sure I could capture "Nothing happened", but the earlier checks were added later. This can safely be removed.

cmd/local.go
shairozan: The only way to handle that would be to have it in the cleanup phase, which is where it was originally, but I wanted to avoid it because of the constant check / write of the file depending on how many models are executed. 

Would just need to move it back to the cleanup phase, since the NonMemModel contains the original path for the submitted model. Check to see if the file exists, then write if not present.

dpastoor: note https://github.com/metrumresearchgroup/babylon/issues/94 

this might be just a misconception on your part. If saving the config, it should go in the model subdir where its executed, not the original dir.

dpastoor:
This is functional, we'll need refactoring, but going to merge in right now to start collapsing down branches for refinement

PR #97, opened 2020-01-02
Feature/running nonmem

dpastoor:


dpastoor:
consolidation of code to be refined

PR #99, opened 2020-01-05
add hashing of data

dpastoor:
This PR accomplishes the following:

1) change config.yaml to config.json as this config is primarily meant for computers and json is a bit faster/more comprehensive parser-wise across languages these days, which outweighs the slight edge to read/editability of yaml

2) implement the data_path and md5 checksum calculation of the model data and provide it to the config.

Using the acop model from testdata:

The resulting config.json is as follows:

```json
{
    "model_name": "acop.mod",
    "model_path": "/Users/devinpastoor/go/src/github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/acop/acop.mod",
    "data_path": "../../../../acop.csv",
    "data_md5": "4ddb44da897c26681d892aa7be99f74b",
    "model_filename": "acop",
    "model_extension": "mod",
    "original_path": "/Users/devinpastoor/go/src/github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/acop",
    "output_dir": "/Users/devinpastoor/go/src/github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/acop/acop",
    "configuration": {
        "overwrite": true,
        "clean_lvl": 2,
        "git": true,
        "babylon_binary": "/data/bbi",
        "save_config": true,
        "output_dir": "{{ .Name }}",
        "threads": 12,
        "nonmem": {
            "nm74": {
                "home": "/opt/NONMEM/nm74",
                "executable": "nmfe74",
                "default": true
            }
        }
    },
    "error": null
}
```

I can additionally launch R from the output dir and confirm the hashes match:

```R
> digest::digest(file = "../../../../acop.csv")
[1] "4ddb44da897c26681d892aa7be99f74b"
```

This will allow for further modifications downstream, such as performing model re-running based on data invalidation, etc.


dpastoor:
after playing around a little more, I've decided to namespace the config just to make sure no conflicts could arise. In addition, added another config field to capture the bbi version used to execute the model

```diff
cat acop/bbi_config.json
{
+    "bbi_version": "2.1.0-alpha.4-2020-01-05T10:05:30-0500",
    "model_name": "acop.mod",
    "model_path": "/Users/devinpastoor/go/src/github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/acop/acop.mod",
    "data_path": "../../../../acop.csv",
    "data_md5": "4ddb44da897c26681d892aa7be99f74b",
    "model_filename": "acop",
    "model_extension": "mod",
    "original_path": "/Users/devinpastoor/go/src/github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/acop",
    "output_dir": "/Users/devinpastoor/go/src/github.com/metrumresearchgroup/babylon/testdata/example-models/nonmem/acop/acop",
    "configuration": {
        "overwrite": true,
        "clean_lvl": 3,
        "git": true,
        "babylon_binary": "/data/bbi",
        "save_config": true,
        "output_dir": "{{ .Name }}",
        "threads": 12,
        "nonmem": {
            "nm74": {
                "home": "/opt/NONMEM/nm74",
                "executable": "nmfe74",
                "default": true
            }
        }
    },
    "error": null
}
```

dpastoor:
at this point, I will consider this code-frozen for incorporation. Other tweaks from usage will be a separate effort.

cmd/local.go
```diff
@@ -199,8 +232,11 @@ func (l LocalModel) Cleanup(channels *turnstile.ChannelMap) {
 	//Gitignore operations
 	createNewGitIgnoreFile(l.Nonmem)
 
+	// this should have been either completed well before, or must at least wait now to complete the hash
+	// before writing out the config
+	l.Nonmem.DataMD5 = <-hashChan
```
shairozan: Wouldn't it be better to have wait group here? Or at least a select with a maximum read count? Just to make sure that we are certain the operation has completed? Here we run the risk of pulling nothing off the channel and continuing on

dpastoor: The point of having that unbuffered channel is such that it will wait until something is on the channel to pull from it though...

> If the capacity is zero or absent, the channel is unbuffered and communication succeeds only when both a sender and receiver are ready.
> If the channel is unbuffered, the sender blocks until the receiver has received the value

```
package main

import (
	"fmt"
	"time"
)

func main() {
	exChan := make(chan string)
	startTime := time.Now()
	go func() {
		sleepTime := 2 * time.Second
		fmt.Printf("in go routine, about to sleep for %s seconds\n\n", sleepTime)
		time.Sleep(sleepTime)
		exChan <- "message on channel"
		fmt.Println("finished in goroutine")
	}()
	fmt.Println("printing before pulling from channel")
	fmt.Println(<-exChan)
	fmt.Println("printing after pulling from channel %s seconds later", time.Since(startTime))

}

```

```
➜ go run channel-wait.go
printing before pulling from channel
in go routine, about to sleep for 2s seconds

finished in goroutine
message on channel
printing after pulling from channel %s seconds later 2.004226966s
```

shairozan: You're right. Didn't pay enough attention to see that it was unbuffered. It's a personal paranoia point of mine to have cancellation contexts and wait groups. I have to deal with this in tests all the time too. Should have paid more attention.

PR #100, opened 2020-01-06
Feature/parallel

shairozan:
Updating to allow for a few flags / inputs to allow parallel operation of nonmem. 

- Defines parallel operation mode
- Updates nonmem command string to use para file
- Generates and writes parafile from parallel flags

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -1,12 +1,15 @@
 package cmd
 
 import (
+	"crypto/md5"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -19,7 +22,6 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"gopkg.in/yaml.v2"
 )
 
 var arguments []string
@@ -137,6 +139,37 @@ func (l LocalModel) Cleanup(channels *turnstile.ChannelMap) {
 	time.Sleep(10 * time.Millisecond)
 	log.Printf("Beginning cleanup phase for model %s\n", l.Nonmem.FileName)
 	fs := afero.NewOsFs()
+	// while the rest of the cleanup is happening, lets also hash the data in the background
+	// so don't have to wait extra time if its on the larger end
+	//Get the lines of the file
+	modelPath := filepath.Join(l.Nonmem.OutputDir, l.Nonmem.Model)
+	sourceLines, err := utils.ReadLines(modelPath)
+
+	if err != nil {
+		log.Printf("error reading model at path: %s, to extract data path: %s\n", modelPath, err)
+	}
+	for _, line := range sourceLines {
+		if strings.Contains(line, "$DATA") {
+			// extract out data path
+			l.Nonmem.DataPath = filepath.Clean(strings.Fields(line)[1])
+		}
+	}
+	hashChan := make(chan string)
+	go func() {
+		f, err := os.Open(l.Nonmem.DataPath)
+		if err != nil {
+			log.Printf("error reading data to hash: %s\n", err)
+			hashChan <- ""
+		}
+		defer f.Close()
+
+		h := md5.New()
+		if _, err := io.Copy(h, f); err != nil {
+			log.Printf("error hashing data: %s\n", err)
+			hashChan <- ""
+		}
+		hashChan <- fmt.Sprintf("%x", h.Sum(nil))
+	}()
 
 	//Magical instructions
 	//TODO: Implement flags for mandatory copy and cleanup exclusions
@@ -166,11 +199,11 @@ func (l LocalModel) Cleanup(channels *turnstile.ChannelMap) {
 		err = utils.WriteLines(source, path.Join(pwi.FilesToCopy.CopyTo, file))
 
 		if err != nil {
-			log.Printf("An erorr occurred while attempting to copy the files: File is %s", v.File)
+			log.Printf("An error occurred while attempting to copy the files: File is %s", v.File)
 			continue
 		}
 
-		v.File = l.Nonmem.OutputDir + "/" + v.File
+		v.File = filepath.Join(l.Nonmem.OutputDir, v.File)
 
 		copied = append(copied, v)
 	}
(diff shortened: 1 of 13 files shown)


cmd/root.go
```diff
@@ -79,6 +79,7 @@ func init() {
 	RootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "verbose output")
 	RootCmd.PersistentFlags().BoolVarP(&debug, "debug", "d", false, "debug mode")
 	RootCmd.PersistentFlags().IntVar(&threads, "threads", 4, "number of threads to execute with")
+	viper.BindPFlag("threads", RootCmd.PersistentFlags().Lookup("threads"))                                               //Update to make sure viper binds to the flag
```
shairozan: This was specifically because I realized that thread settings set from the command line were not overriding data in the config file from `bbi init`. Probably need to actually go through and make sure verbose, debug etc are also bound to viper

configlib/config.go
```diff
@@ -29,6 +30,13 @@ type NonMemDetail struct {
 	Default    bool   `yaml:"default"`
 }
 
+type ParallelConfig struct {
+	Parallel    bool   `yaml:"parallel" json:"parallel,omitempty"`
+	Nodes       int    `yaml:"nodes" json:"nodes,omitempty"`
+	MPIExecPath string `yaml:"mpiExecPath" json:"mpiExecPath,omitempty"`
+	Timeout     int    `yaml:"timeout" jason:"timeout,omitempty"`
```
dpastoor: would like to meet jason one day

cmd/nonmem.go
```diff
@@ -220,6 +261,47 @@ func generateScript(fileTemplate string, l NonMemModel) ([]byte, error) {
...
+	contentLines := strings.Split(string(contentBytes), "\n")
+
+	return utils.WriteLines(contentLines, path.Join(l.OutputDir, paraFileName))
+
+}
+
+func generateParaFile(l NonMemModel) ([]byte, error) {
+	nmp := nonmemParallelDirective{
+		TotalNodes:        l.Configuration.Parallel.Nodes,
+		HeadNodes:         l.Configuration.Parallel.Nodes - (l.Configuration.Parallel.Nodes - 1),
```
dpastoor: this logic will just always come to 1? why not just code it as `1`?

cmd/nonmem.go
```diff
@@ -49,6 +49,29 @@ const sgeExecutionTemplate string = `#!/bin/bash
 {{ .Command }}
 `
 
+//Parse type 2 refers to evenly load balanced work
+//Transfer Type 1 refers to MPI
+//TIMEOUTI 100 means wait 100 seconds for node to become available
+//TIMEOUT 10 means wait 10 seconds for work to complete -> Should default to 10
```
dpastoor: many models can take days or even weeks to run. It would be best to have a default of no timeout (can timeout set to 0 or maybe a negative number do that?)

cmd/local.go
```diff
@@ -236,6 +245,11 @@ func local(cmd *cobra.Command, args []string) {
 		viper.Debug()
 	}
 
+	//Check for parallel mode provided with more nodes than CPU Cores
+	if viper.GetInt("parallel.nodes") > runtime.NumCPU() || viper.GetInt("parallel.nodes")*viper.GetInt("threads") > runtime.NumCPU() {
```
dpastoor: I don't think this is valid, as I _believe_ you can run in parallel across multiple nodes. It doesn't happen that often but we should be careful around forcing something to fail

dpastoor:
One other very important thing to consider, is when submitting via the grid engine, the parallel operation should properly take up the number of slots needed, which should match the number of execution nodes.

cmd/local.go
shairozan: Yeah this has already been removed locally. Made sense when I was putting everything together, but not as I thought it through and started trying grid execution

cmd/nonmem.go
shairozan: This is primarily from the documentation on Nonmem:

```
; NODES=number of nodes (that is process, whether cores or computers)
; SINGLE node: NODES=1
; MULTI node (node means process, whether cores or computers): NODES>1
; WORKER node: NODES=0
;
; parse_num=number of subjects to give to each node
; parse_type=0, give each node parse_num subjects
; parse_type=1, evenly distribute numbers of subjects among available nodes
; parse_type=2, load balance among nodes
; parse_type=3, assign subjects to nodes based on idranges
; parse_type=4, load balance among nodes, taking into account loading time.
; This setting of parse_type will assess ideal number of nodes.
; If loading time too costly, will eventually revert to single CPU mode. ;
; timeouti=seconds to wait for node to start. if not started in time,
; deassign node, and give its load to next worker, until next iteration
; timeout=minutes to wait for node to compelte. if not completed by then, ; deassign node, and have manager complete it.
; paraprint=1 print to console the parallel computing process. Can be
; modified at run-time with ctrl-B toggle.
; Regardless of paraprint setting, <control_stream>.log always records
; parallelization progress.
```
I need to try setting it to 0 and seeing what it does  because the nonmem docs don't specifically say anything about unlimited timeout in parallel execution mode

shairozan: So the job still finishes with a defined timeout of 0. Not sure if it's exerting extraneous pressure on the manager mentioned above, but if it works fine we can basically always set this to 0 and remove the parameter altogether

shairozan: The manager according to those docs is the thread invoking the nonmem command. So from what I understand, if the parallel node doesn't finish the work in the time frame specified, it's handed back to the manager to do the work. So to some degree it ceases to be parallel.

> The manager process is the user’s process that runs the nmfe74 script, reads the control stream file, executes NMTRAN, and runs the main NONMEM process. The worker process is NONMEM in worker mode, not taking any input from the user, only from the manager NONMEM process.

shairozan: Also using a negative number breaks nonmem compilation:
```
 * THE MAXIMUM NUMBER OF WARNINGS OF ONE OR MORE TYPES WAS REACHED.
 IT IS POSSIBLE THAT SOME WARNING MESSAGES WERE SUPPRESSED.
Recompiling certain components
  TIMEOUT MUST BE INTEGER BETWEEN (0,2147483647)
```

2147483647 looks like a sane value to use. I mean that's like 4000 years

configlib/config.go
shairozan: Not today son. He has left the building

cmd/nonmem.go
shairozan: Yeah. I wasn't exactly sure if I was going to have some other operation when I originally wrote it, but every single pnm file I could find always just had 1, so it does just make sense to leave it at 1. Was originally also going to do the math via functions mapped to the template and realized that was stupidly overkill for arithmetic that could be done in language.

dpastoor:
first tests looking good - update on some notes here: https://github.com/metrumresearchgroup/babylon/issues/95

shairozan:
Just updated the branch with a section to handle the files / directories listed.

shairozan:
Also updated with a new commit allowing the user to specify `--parafile /path/to/parafile.pnm`, and we will simply read those lines and write into the directory as opposed to the generated file we would normally use.

Issue #93, opened 2019-12-31
overwrite does not seem to do anything


dpastoor:
![image](https://user-images.githubusercontent.com/3196313/71625332-36ff5700-2bb5-11ea-952a-5136c5114d27.png)

notice it takes 16 seconds (time to run models).

The shell script does not seem to be modified, but all others have new timestamps, suggesting it is re-running

![image](https://user-images.githubusercontent.com/3196313/71625355-5ac29d00-2bb5-11ea-8ccf-5f39577743a9.png)

dpastoor:
update: in 2.1.0-alpha.2 the following now occurs:

```
devinp@ip-10-0-1-7:~/nonmem-test-master/model$ bbi nonmem run local 103.mod --overwrite=false
2020/01/04 20:25:55 Configuration file successfully loaded from /data/home/devinp/nonmem-test-master/model/babylon.yml
2020/01/04 20:25:55 A total of 1 models have been located for work
2020/01/04 20:25:55 1 models successfully completed initial setup phase.
2020/01/04 20:25:55 Beginning local work phase for 103
2020/01/04 20:25:55 Exit code was 106, details were exit status 106
2020/01/04 20:25:55 output details were: File /data/home/devinp/nonmem-test-master/model/103/103.mod does not exist.
rm: cannot remove 'FSUBS_MU.F90': No such file or directory
2020/01/04 20:25:55 2 errors were experienced during the run
2020/01/04 20:25:55 Errors were experienced while running model 103.mod. Details are The target directory, /data/home/devinp/nonmem-test-master/model/103 already, exists, but we are configured to not overwrite. Invalid configuration / run state
2020/01/04 20:25:55 Errors were experienced while running model 103.mod. Details are Running the programmatic shell script caused an error
1 models completed in 59.321617ms
```

This looks like its still attempting to do a full clean of the directories and does not break out of that operation if the model doesn't run

> rm: cannot remove 'FSUBS_MU.F90': No such file or directory

shairozan:
This is more of an issue with turnstile where I have to address cancellation of phases after the error occurs. In this scenario, we exit the prep phase and mark the error, but the other phases still attempt to continue.

shairozan:
I've got this successfully implemented and tested in turnstile. Just need to work through making the changes on a branch here and make sure everything is still working as expected.

shairozan:
Looks like I've got this working in branch issue/93

shairozan:
[...]

PR #101, opened 2020-01-08
Updating to use new cancel functionality included in turnstile

shairozan:
Addressing issue #93 by implementing cancel channels for turnstile


shairozan:
Also included logic for handling overwrite = false to not just give up unless the directory contains copiable files. Addresses #102

PR #106, opened 2020-01-14
Building platform for processing provided config 

shairozan:
Mechanism of handling a statically provided config without interfering with SGE process


shairozan:
@dpastoor Can you test this and look for any shortcomings? I drew through the process several times as to make sure it wouldn't interfere with SGE.

# Summary of Changes

* We don't hunt for configs anymore. It's either in the original model directory or not at all
* At the beginning of both the SGE and Local execution we check to see if the config flag was set, if so we load that config. 


# Implications
* Because of this, If no configuration is either loaded from the model or via `--config` flag, the application will die with a fatal log indicating that no configuration file was found in path. 
    * Based on previous discussions and the need to use init to at least locate nonmem, I don't see this as an issue as that seemed to be the direction we were going. 
* The locateAndLoadConfigs method called in the creation of all new NonMemModels doesn't really execute because viper will already register that a configuration file is loaded. So we don't have to worry about double-reading or multiple reading of config files
* SGE's path uses this same process at the beginning to:
    * Write its configs to the default location
    * Read that config back in. This prevents anything from trying to reload the viper configs.

Issue #108, opened 2020-01-14
update init to set default to true if only one nonmem install present


dpastoor:
for simple user cases where only one version of nonmem installed could be nice if init just set that default to true

shairozan:
This is available in the master branch and most recent release

PR #111, opened 2020-01-20
Updating to default to true if only one nonmem location is found

shairozan:
Quick fix to address #108


shairozan:
@dpastoor 

I updated the init path such that if we only have one directory located during init phase, we'll set it to true. Works fine locally on the Mac. Tested both pathways of only having one, then creating another directory that would match to make sure that greater than one still has false for all entries.

PR #113, opened 2020-01-20
Updating with a randomized timer for local mode execution

shairozan:
* Random is initialized off of the timer during root init
* `--delay` flag introduced with default value of 0
* If delay value in configuration is greater than default:
    * A random integer between 1 and the provided delay is selected
        * This avoids synchronization. If you just specify a delay of 4 and 4 seconds is selected, all of the jobs will begin execution after waiting 4 seconds. This would still yield the IO problems. Just on a delay
        * If debug is set, it will display the random seconds elected for delay

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -52,7 +52,7 @@ func (l LocalModel) Prepare(channels *turnstile.ChannelMap) {
 		randomizedTimer := randomFloat(1, l.Nonmem.Configuration.Delay)
 
 		if l.Nonmem.Configuration.Debug {
-			log.Printf("Random delay of %f seconds introduced for model %s", randomizedTimer, l.Nonmem.FileName)
+			log.Infof("Random delay of %f seconds introduced for model %s", randomizedTimer, l.Nonmem.FileName)
 		}
 
 		time.Sleep(time.Duration(randomizedTimer) * time.Second)
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -45,6 +45,18 @@ func (l LocalModel) CancellationChannel() chan bool {
 //Prepare is basically the old EstimateModel function. Responsible for creating directories and preparation.
 func (l LocalModel) Prepare(channels *turnstile.ChannelMap) {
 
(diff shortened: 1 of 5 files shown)


cmd/root.go
```diff
@@ -78,10 +84,11 @@ func init() {
 	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/babylon.yaml)")
 	RootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "verbose output")
 	RootCmd.PersistentFlags().BoolVarP(&debug, "debug", "d", false, "debug mode")
+	viper.BindPFlag("debug", RootCmd.PersistentFlags().Lookup("debug")) //Bind Debug to viper
```
shairozan: This was updated because the debug values in the object weren't represented. Even if you had set this to true, it wouldn't show up when viper is marshalled into the struct on the job.

shairozan:
@dpastoor 

When you get a chance, take a look at this, primarily to see if you have any issues with the assignment of random values. I wanted to avoid carte-blanche taking the `--delay` value provided by the user and just waiting for that value, as it would just cause a synchronization issue.

shairozan:
This way it also gets written into the config, so when the local execution branch begins on an SGE worker, it'll read it and say "I have to wait X seconds before starting work"

cmd/local.go
```diff
@@ -56,7 +69,7 @@ func (l LocalModel) Prepare(channels *turnstile.ChannelMap) {
 		//If so are we configured to overwrite?
 		if l.Nonmem.Configuration.Overwrite {
 			log.Debugf("%s Removing directory %s", l.Nonmem.LogIdentifier(), l.Nonmem.OutputDir)
-			err := RemoveContentsAndDir(l.Nonmem.OutputDir)
+			err := fs.RemoveAll(l.Nonmem.OutputDir)
```
shairozan: Moving back to afero as the target for this since this method was created during troubleshooting for the SGE output stream issue.

PR #114, opened 2020-01-20
Adding configs to write SGE output to the output dir of the model

shairozan:
Updating to place a single, converged output file from SGE into the output dir of the directory.

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -5,8 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	log "github.com/sirupsen/logrus"
 	"io"
-	"log"
 	"os/exec"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


shairozan:
It's basically a static output file now called sge.output

```
darrellb@ip-10-0-1-181:/data$ cat 240/001/sge.output
2020/01/20 20:36:59 Configuration file successfully loaded from /data/240/babylon.yml
2020/01/20 20:36:59 A total of 1 models have been located for work
2020/01/20 20:36:59 1 models successfully completed initial setup phase.
2020/01/20 20:36:59 We could not locate or read the mod file (001.mod) indicated to locate output files. As such none will be included in copy / delete operations
2020/01/20 20:36:59 No Nonmem output files detected in /data/240/001. Good to continue
2020/01/20 20:36:59 Beginning local work phase for 001
2020/01/20 20:37:12 Beginning cleanup phase for model 001
1 models completed in 13.076669407s
```

But it is generating in both SGE and parallel SGE execution now

shairozan:
Updated filename to be sge.output.$JOB_ID so that it won't collide across multiple runs. @dpastoor Do we want to add sge.output.* to the gitignore? Or is this something we would want to let people commit to the CVS?

shairozan:
It might also make sense to create a directory called sge_output in the model output dir, placing the sge output texts into those directories. That way you could just gitignore the entire directory and keep the files separate from the model data

shairozan:
And have it working. What I did was the following:

- SGE cleanup phase was previously empty
- Added three Steps to Cleanup:
    - Manually set viper's `overwrite` value to false
    - Manually set the `saveConfig` value to false
    - Write the babylon config back out
   
This keeps the second stage of execution (The local phase) from trying to attempt overwrite operations. As long as there are no Nonmem output files in the directory (Which would be nigh impossible due to the SGE phase still dropping and creating a new directory) it will continue as per normal. 

Setting `saveConfig` to false also keeps the local execution phases from trying to re-write to the config over and over again. The only sane way I can think of is with a global.

This also relieves the conflict with SGE over NFS locks in the directory. The only thing I'd like to resolve would be how to keep

PR #117, opened 2020-01-27
Squashing config write if errors exist

shairozan:
Updating both execution paths to prevent further action if errors are detected


shairozan:
If any errors generate failures before turnstile starts (IE Couldn't find model), an error message is displayed and execution halts. Keeps the app from writing the default config in the event of failures

PR #118, opened 2020-01-27
Don't overwrite config in model dir if config option specified

shairozan:
If the `--config` parameter is provided, we shouldn't just overwrite the configuration in the model directory.

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -63,7 +63,7 @@ func (l LocalModel) Prepare(channels *turnstile.ChannelMap) {
 
 	fs := afero.NewOsFs()
 
-	log.Debugf("%s Overwrite is currently set to %t", l.Nonmem.LogIdentifier(), viper.GetBool("debug"))
+	log.Debugf("%s Overwrite is currently set to %t", l.Nonmem.LogIdentifier(), l.Nonmem.Configuration.Overwrite)
 	//Does output directory exist?
 	if ok, _ := afero.DirExists(fs, l.Nonmem.OutputDir); ok {
 		//If so are we configured to overwrite?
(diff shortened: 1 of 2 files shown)


cmd/local.go
```diff
@@ -385,10 +385,9 @@ func local(cmd *cobra.Command, args []string) {
 		time.Sleep(5 * time.Millisecond)
 	}
 
-	if len(lo.Models) > 0 {
-		if viper.GetBool("saveConfig") {
-			configlib.SaveConfig(lo.Models[0].Nonmem.OriginalPath)
-		}
+	//If set to save config and the config value is not populated
+	if viper.GetBool("saveConfig") && len(viper.GetString("config")) == 0 {
```
shairozan: The `config` value is never serialized down to the `l.Nonmem.Configuration` struct on purpose. Conceptually this completely avoids potential loops for configurations being loaded. As such we reach directly to viper to see what the value is. If the value is not am empty string, even if `saveConfig` we won't save the file.

cmd/sge.go
```diff
@@ -375,10 +375,24 @@ func generateBabylonScript(fileTemplate string, l NonMemModel) ([]byte, error) {
 		Command          string
 	}
 
+	commandComponents := []string{
```
shairozan: No logical change here, just generating the command from string slices to make it cleaner.

shairozan:
So this behavior is unavoidable for the way we have SGE working. Once you specify a configuration, it gets loaded into viper, the SGE cleanup phase sets some values in viper and writes the file out to the original model directory for local execution from the grid. 

@dpastoor Do we have any issues with this still writing the config in an SGE sense? If not, we can update the command to use the config flag, but we would also have to do some work dealing with 
* What file is loaded
    * Do we copy the provided one and then change some values to make sure overwrite doesn't happen?
    * Write that file and use it in the shell script

shairozan:
This was actually addressed by the re-write of config storage. We never overwrite the source yaml anymore, but rather render the merged configuration into the output directory for the models.

PR #146, opened 2020-03-17
Cleanup/flag case consolidation and cleanup

shairozan:


shairozan:
@dpastoor 

This is purely a cleanup of documentation branch. No code changes. I went through and removed the global flags from each subsequent child command since it keeps the flags from just being repeated page after page.

shairozan:
Added a fix for #150 into this branch.

docs/bbi/nonmem/run/local/local.md
```diff
@@ -9,30 +9,16 @@ Specifies to run the targeted model on the local machine
 ### Options
 
 ```
-      --cacheDir string       directory path for cache of nonmem executables for NM7.4+
-      --cacheExe string       name of executable stored in cache
```
Dreznel: Do all of these options no longer exist?

shairozan: They're not implemented anymore. They were from the original incarnation of Babylon, but are not actually implemented anywhere now.

PR #148, opened 2020-03-18
feature: initial pass at post execution. Will eventually move it to a…

shairozan:
…n interface based implementation


dpastoor:
FYI i haven't forgotten about this - but don't currently have permissions to set this up in our metrumrg space. I'd like to get a dev slack sandbox set up for testing this and other tools that interact with slack.

shairozan:
It's all good. I used a slack space that I administer to do the same, so it would make sense to have a similar setup here.

shairozan:
We do also already have a suite of tests for this in Babylon test in a separate branch

PR #153, opened 2020-03-27
fix: updating goreleaser to make sure it'll build windows binaries on release

shairozan:


shairozan:
No application code changes. Just goreleaser configs, verified with goreleaser in snapshot mode locally.

Noticed that no windows binaries were compiled for the release. Even though the default goreleaser configs should have built for all three, once we had to specify a new `builds` section, it needed to have all three defined since its defaults are only windows / linux.

Issue #156, opened 2020-04-03
Mulitple model submission --overwrite flag is not passed through


seth127:
What seems to be happening is that, when you pass multiple model files as arguments, the `--overwrite` CLI flag is only passed through to the _first_ of those models.

Example:
```
$ cp 1.ctl 2.ctl

$ /data/apps/bbi nonmem run sge 1.ctl 2.ctl

$ /data/apps/bbi nonmem run sge 1.ctl --overwrite
INFO[0000] Successfully loaded default configuration from /data/home/sethg/rbabylon/tests/testthat/model-examples-bbi/babylon.yaml
INFO[0000] Beginning Local Path
INFO[0000] A total of 1 models have been located for work
INFO[0000] [1] Beginning SGE work phase
1 models completed in 10.174348ms

$ /data/apps/bbi nonmem run sge 1.ctl 2.ctl --overwrite
INFO[0000] Successfully loaded default configuration from /data/home/sethg/rbabylon/tests/testthat/model-examples-bbi/babylon.yaml
INFO[0000] Beginning Local Path
INFO[0000] A total of 2 models have been located for work
INFO[0000] [1] Beginning SGE work phase
ERRO[0000] 1 errors were experienced during the run
ERRO[0000] Errors were experienced while running model 2. Details are The target directory, /data/home/sethg/rbabylon/tests/testthat/model-examples-bbi/2 already exist, but we are configured not to overwrite. Invalid configuration / run state
2 models completed in 10.180546ms
```

I originally thought maybe _none_ of the flags were passed through, but I tried it with `threads=4` and then looked at `1/bbi_config.json` and `2/bbi_config.json` and they both had `threads: 4` in them. (When I didn't pass the flag, they both had `threads: 2` in them.

shairozan:
This isn't actually an issue with it not being passed through, but rather an issue of concurrency. We use a pointer to the configuration object as a child to the Nonmem objects, so basically they're all _sharing_ a configuration as opposed to copying it. 

What is happening, however, is that the values are manipulated during SGE submission such that the values of the babylon.yaml file written into the output dir are modified. Three things are changed:

* Output = false
* CreateChildDirs = false
* SaveConfig = false

[...]

PR #157, opened 2020-04-03
fix: dereferencing config before writing yaml

shairozan:
Addresses #156 

The overwrites values are correct in initial collection, injection into the model, but periodically could observe the value as false during prepare phase. Realized that it was because the 

```go
func WriteViperConfig(path string, sge bool, config *Config) error {

	//Dereference to avoid touching the original values for multiple run targets
	lc := *config

	if sge {
		//Set the config to overwrite false and re-write config. This ensures that the local phase will not deal with io contention
		//around the SGE output streams
		log.Debug("Updating babylon config to overwrite=false. This avoids IO contention with the grid engine for the next execution round")
		lc.Overwrite = false
		lc.SaveConfig = false
		lc.Local.CreateChildDirs = false
	}

	log.Debugf("Requested save of config file %s", path)

	err := lc.RenderYamlToFile(path)

	if err != nil {
		return err
	}

	return nil
}
```

block previously used a pointer directly (because it was available in the model) and acted directly on it. Now we dereference and manipulate a local copy to avoid unexpected action

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 .idea
 dist
 babylon.log
+babylon.yaml
 !.idea
--- a/configlib/config.go
+++ b/configlib/config.go
@@ -113,19 +113,22 @@ func SaveConfig(configpath string) {
 }
 
 func WriteViperConfig(path string, sge bool, config *Config) error {
-	if sge {
 
+	//Dereference to avoid touching the original values for multiple run targets
+	lc := *config
+
+	if sge {
 		//Set the config to overwrite false and re-write config. This ensures that the local phase will not deal with io contention
 		//around the SGE output streams
 		log.Debug("Updating babylon config to overwrite=false. This avoids IO contention with the grid engine for the next execution round")
-		config.Overwrite = false
-		config.SaveConfig = false
-		config.Local.CreateChildDirs = false
+		lc.Overwrite = false
+		lc.SaveConfig = false
+		lc.Local.CreateChildDirs = false
 	}
 
 	log.Debugf("Requested save of config file %s", path)
 
-	err := config.RenderYamlToFile(path)
+	err := lc.RenderYamlToFile(path)
 
 	if err != nil {
 		return err


shairozan:
The rest of the application uses pointers to the nonmem model and pointers to the Config so that, in the concurrent context of Turnstile, the config exists once and isn't shadowed all over the place. As such, everything for the nonmem models and the config are based purely on the pointer reference. 

We could dereference in the scope of the turnstile Prepare implementation before the method is called, but that would mean each implementation (SGE, Local) would have to dereference in their own implementations. 

To me, since that would require adjustments in each implementation, it's easier and safer to just dereference in the method, since that's the only place where we want to de-couple, and it's only for that point in time as we're making changes to the values to write them down for the next execution stage.

bkyoung:
My 2 cents: I don't know that "clarity" is so much the issue or concern in this situation.  Dereferencing a pointer is just ... that.  ¯\_(ツ)_/¯  The larger question is are we passing a pointer because we're communicating by sharing memory?  If it's being done for any other reason, it's probably being done for the wrong reason - unless you can show me the before and after profiler results that prove otherwise.  Recall that the (only valid) reason for using a pointer is for communicating by sharing memory across a program boundary (like function to function calls, or across go routines, to name a couple of common ones).  Also recall that a struct is not a "concurrency safe" data structure, meaning that the outcome is undefined if two different go routines updated the same struct field simultaneously.

If we're "doing it wrong", does that mean we should go change it everywhere?  My inclination is to leave it unless it shows itself to be problematic in some way.  If there is time and desire, it could be a worthwhile exercise to do the math and figure out how much memory the Config struct is using, and profile things to see what impact there would be in passing the struct by value rather than passing around the pointer value of the pointer to a struct (man, that's a mouthful).  Might be worth a larger think on adding to CI/CD at some point to keep those profiler results as a record over time.   But, without that info, any "optimizations" are guesses and wasted effort.

dpastoor:
Hmm ok this was the concern I had with the pointer semantics early around unintended global effects. Implementation wise this makes sense for this fix, but worth taking a pause if there are other places this type of scenario could come up and get comprehensive tests in place. There will always be the need to keep in mind effect of switching a value from the config on all other models.

I can see places where having shared state can be important, especially around orchestration and error management, but we need to be extra careful with testing.

I also appreciate this is extra tricky as all the sge stuff is more difficult to test as we either mock or can only run in specific environments.

We'll grow out more tests in rbabylon and see if we can get some more solid scenarios to make sure we're extra cognizant of

PR #158, opened 2020-04-04
refactor: removing the configuration as a pointer

shairozan:
Closed the previous pull request. This branch removes the configuration as pointer altogether.

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -297,7 +297,7 @@ func local(cmd *cobra.Command, args []string) {
 	lo := localOperation{}
 
 	log.Debug("Locating models from arguments")
-	localmodels, err := localModelsFromArguments(args, &config)
+	localmodels, err := localModelsFromArguments(args, config)
 
 	if err != nil {
 		log.Fatalf("An error occurred during model processing: %s", err)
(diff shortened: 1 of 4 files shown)


shairozan:
All tests were previously passing and I've run it manually to confirm that it's resulting in the same behavior the previous branch did. Just different approach. Going to have Seth test it on Monday

shairozan:
```go
if l.Nonmem.Configuration.NMQual {

		//Set parallelism to true
		l.Nonmem.Configuration.Parallel = true
		//TODO: What about the other parallel components?

		//If we can locate the key
		if selected, ok := l.Nonmem.Configuration.Nonmem[l.Nonmem.Configuration.NMVersion]; ok {
			//Is it not set to nmqual?
			if !selected.Nmqual {
				RecordConcurrentError(l.Nonmem.FileName, "Invalid nonmem / nmqual configuration", errors.New("NMQual was selected, but the selected nmversion does not support nmqual"), channels, l.Cancel)
				return
			}
		}
	}
```

Had me slightly concerned as it was the only place that we reached into the configuration to change a value from the execution perspective. I have verified, however, that both the generated Babylon yaml file and the JSON file both contain a `parallel` value of true after execution.

dpastoor:
@shairozan Good to merge and tag so can use this in a build?

shairozan:
Yeah it's good. Was waiting for Seth to update the issue confirming it worked. Been working on testing for Babylon stuff. I'll go ahead and merge into dev and create a new release

shairozan:
@dpastoor v2.1.1-rc1 is available now.

PR #159, opened 2020-04-08
Fixing failure to hash source model for NMQual

shairozan:
# Source Problem
When running with`--nmqual`, errors are generated in the output indicating the source model could not be hashed

# Reason
During operation in the local branch, we copy the model, rename it to a `*.ctl`. This is because the NMQual operation requires a filename ending with CTL. At this point, however, we have no stored reference back to the original `.mod` or `.whatever` file.  So the hash tries to find a `.ctl` file in the original source directory (which may not exist). 

# Fix
We already had an `OriginalPath` member of the `NonMemModel` struct, so I added an `OriginalModel` field, which is populated when all of the models are created from arguments. During hashing operations we always go back to the `OriginalPath`/`OriginalModel` to accurately get a model hash.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,6 +22,7 @@
         MPIEXEC_PATH: /usr/bin/mpiexec
         NONMEMROOT: /opt/NONMEM
         NMVERSION: nm74gf
+        NMQUAL: true
         NONMEM_LICENSE:
           from_secret: NONMEM_LICENSE
       image: nmtest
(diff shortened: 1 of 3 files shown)


shairozan:
Specifically deals with issue found in #159 located during some manual tests

shairozan:
We do have a test for it in [Babylontest](https://github-drone.metrumrg.com/metrumresearchgroup/babylon/299/1/4) now:

PR #162, opened 2020-04-16
Execution of testing on SGE Test Runner

shairozan:
This branch makes changes to the drone configuration for babylon

* Moved to Execution Pipeline
* Anchored variable sets for SGE / SGE Pathing
* Mounting of volumes (including `/data` from host)
* Inclusion of Pandoc to convert to PDF and render final versions

It also adds / changes the following behaviors in Babylon:

* Adds a `grid_name_prefix` flag to the SGE command. This basically allows a user to prepend any value to the name we select when submitting to the grid. This is used to "cordon" off one run from another by allowing the test suite to know exactly what jobs to look for on the grid
* Adds the `grid_name_prefix` to the config object
* Updates the queue submission process to:
    * Not just simply concatenate "Run_" + model.Identifier
    * Implements a template operation to build:
         * `{{your_prefix}}_Run_{{ model.Identifier}}` if the prefix is not empty
         * `Run_{{model.identifier}}` if prefix is empty


validation.json
```diff
@@ -23,9 +23,6 @@
     {
       "name" : "Babylon should allow users to run NonMem jobs on the Grid",
       "tags" : [
-        "qsub",
-        "qstat",
```
shairozan, line 27: Removing manual tests from validation spec

shairozan:
Babylon Pipelines

This is the structure we'll be going for. For all git events, Babylon will run the testing suite. There are then two pipelines that will depend on it. The first will run on the same SGE runner, which will do the validation. This is primarily because the instance is the only thing allowed to access the S3 bucket by policy. Then in parallel, in one of the drone workers using a docker pipeline, we will execute the goreleaser work to publish the version

shairozan:
Screen Shot 2020-04-16 at 10 53 00 AM
Looks good so far. Have a split execution across the three pipelines

shairozan:
So right now the drone execution agents are running with a max capacity of 1. If I can fix a couple of things, I think I can safely remove this restriction.

1. Execution Dir Location:  Currently the work is done scenario by scenario under /data/working. As things are, this would overlap between executions causing failures about files not being there when expected. To resolve this, the ROOT_EXECUTION_DIR will be set to `/data/${DRONE_BUILD_NUMBER}/working` An additional pipeline after testing will also be required to attempt to clean that up in the event of failure
2. Grid engine naming: Currently when submitting jobs to the grid, we submit with the name of `Run_{model_name}`. It's a simple string concatenation, but we could move this to a template with an additional SGE flag to provide prefixing. 

What I'm thinking of is a template of `{{ if .Prefix}} {{ .Prefix}}_{{.Filename}}{{else}}{{.Filename}}{{end}}`
Prefix would be set from a `--grid_name_prefix` flag that could be bound from the environment or command flag. This would allow drone to set the grid name prefix to the ${DRONE_BUILD_NUMBER} variable.

This way, each build would have specific job names. Meaning we could use gogridengine in babylontest to look for the status of the specific job name as opposed to any jobs in `qw` state.

shairozan:
So isolating the workspaces via drone build number worked fine. Need to add cleanup steps and then handle the grid submission issue in Babylon and Babylon test

shairozan:
The cleanup is working. I had to do an inline attempt in the test step and then add a cleanup step conditionally active on step failure.

shairozan:
Now to tackle the templated based job submission name and prefix flag

shairozan:
Ok forgot an issue surrounding name of jobs submitted to the grid. Originally was just pretending the drone build number (a number) to the job name, but SGE won't allow that. Updated the drone config to make it drone_${DRONE_BUILD_NUMBER} and SGE is executing again. 

Next I need to make sure that:

* BBI is being built and moved into `/data/{$DRONE_BUILD_NUMBER}/apps`
* Path is updated to that

That'll guarantee binary isolation. So at that point:

* No two jobs should share a BBI binary
* No two jobs should share space on the file system
* All jobs are specifically locatable by scenario prefix. 

Will try scaling to 4 jobs on the drone host and see how it goes o.o

shairozan:
I've got this now successfully running more than one at a time, and have the host configured for 4 concurrent jobs now. There were some machine sympathies I had to add to a few tests in babylontest where we were just trying to fan out across too many jobs for the cores on the box. After that, the tests are looking a bit more stable.

PR #173, opened 2020-08-07
adding condition number to summary

seth127:
Pulling out the actual condition number instead of just surfacing the heuristic of whether it's > 1000.

### DIFF
--- a/parsers/nmparser/parse_lst_file_test.go
+++ b/parsers/nmparser/parse_lst_file_test.go
@@ -227,11 +227,36 @@ func TestConditionNumber(t *testing.T) {
 		// 	largeConditionNumber: Undefined,
 		// 	context:              "not set",
 		// },
+
+		{
+			lines: []string{
+				"                                                                                                                                   ",
+				" ************************************************************************************************************************          ",
+				" ********************                                                                                ********************          ",
+				" ********************               OBJECTIVE FUNCTION EVALUATION BY IMPORTANCE SAMPLING             ********************          ",
+				" ********************                   EIGENVALUES OF COR MATRIX OF ESTIMATE (RSR)                  ********************          ",
+				" ********************                                                                                ********************          ",
+				" ************************************************************************************************************************          ",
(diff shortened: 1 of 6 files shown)


seth127:
Per @shairozan 's request, here is the unit test for this
`go test github.com/metrumresearchgroup/babylon/parsers/nmparser -v -run TestConditionNumber`

parsers/nmparser/parse_lst_file_test.go
```diff
@@ -226,12 +227,38 @@ func TestSetLargeConditionNumber(t *testing.T) {
 		// 	largeConditionNumber: Undefined,
 		// 	context:              "not set",
 		// },
+
+		{
+			lines: []string{
```
shairozan, line 232: More of a comment than anything, you can store multiline text components with the backtick operator: 

```go
things := `
I'm a multi
line string
`
```

shairozan:
It looks like you built and committed the main binary. I'd update .gitignore and remove the main binary from the tree so we can avoid committing binaries to the repo

shairozan:
https://medium.com/better-programming/how-to-remove-committed-files-from-git-version-control-b6533b8f9044

Point of reference on how to remove a file from the cache. Won't actually remove the file, just remove it from the git tree

seth127:
@shairozan ok, I gitignore'd it so we won't do it again. @dpastoor and I had a convo about this today but long story short: 

* you had done this same thing back in Dec. 2019 so the binary is actually on the master branch right now. 
* I rebuilt it and committed it by accident (which registered as changes to the existing binary)
* then I realized that and deleted it and did `git rm main`

So the net result it that it looks like it was there and I deleted it. Which is guess is fine for now. We talked about trying to purge it from the history, but that's a fair amount of history so we're gonna leave it alone for now. Are you cool with that?

shairozan:
We don't really care that it's in the history, just that it's not part of future branches, so we should be fine then.

PR #174, opened 2020-08-11
Refactor/cov cor out

seth127:
Refactoring the `bbi nonmem summary` call to make the following changes:

* `bbi nonmem summary` will no longer parse the `.cov` and `.cor` files
  * Take the `.cov` and `.cor` output out of the summary json
* Create new command to extract the information from those files into a separate json.

### DIFF
--- a/cmd/covcor.go
+++ b/cmd/covcor.go
@@ -1,0 +1,53 @@
+// Copyright © 2016 Devin Pastoor <devin.pastoor@gmail.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	parser "github.com/metrumresearchgroup/babylon/parsers/nmparser"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/spf13/cobra"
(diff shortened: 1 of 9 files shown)


seth127:
@dpastoor @shairozan I've implemented this, but do you have any thoughts on how we want to test it at this point? I tested manually with the following (on a model that I copied down from Ana's project):
```
bbi nonmem summary 300/300.lst         # works as it previously did
bbi nonmem summary 300/300.lst --json  # prints json without cov and cor
bbi nonmem covcor  300/300             # prints json of only cov and cor
bbi nonmem covcor  300/300 --json      # prints json of only cov and cor
```

I don't see any model output checked into the repo anywhere at this point. Should I just check some in (from some sample model) and run those commands on them?

I know we have talked about refactoring a lot of these tests, but I'm not sure how imminent that is...

shairozan:
One thing you could do is manually provide models into tests as multi line strings, or they can be added into the babylontest repo. It's publicly available and can be pulled down at will. I would lean towards testing the application functionality from the user's standpoint, similar to how we're doing in babylontest. 

As such, we would want to add scenarios with models we're familiar with, run the application, collect the standard out and verify that it's what we would expect. Then you may also want to do that across multiple versions of nonmem just to be sure.

shairozan:
A few inputs:

* I like the use of flags, but not in tests. Tests are meant to be automated, and there's not a really sane way to force drone to. Because the rest of the test suite is written around ENV variables since they're more easily controlled by drone, my recommendation would be to use ENV configuration and set defaults early on. Not a requirement, but given the nature of the application and how it is used, it may make more sense
* When you create the pull request, make sure you pick a requestor for review if it's ready. I just realized I wasn't on the PR.

seth127:
Added some more tests for this in https://github.com/metrumresearchgroup/babylontest/pull/16
```
$ ROOT_EXECUTION_DIR=/tmp/ go test -v -run TestCov
=== RUN   TestCovCorHappyPath
--- PASS: TestCovCorHappyPath (0.03s)
=== RUN   TestCovCorErrors
--- PASS: TestCovCorErrors (0.02s)
PASS
ok      github.com/metrumresearchgroup/babylontest      0.385s
```

PR #189, opened 2020-09-19
fix shrinkage extraction when more than 50 etas in model

dpastoor:


dpastoor:
@timwaterhouse let me know if you need a fix this weekend, and we can push a release candidate tag. Otherwise, @seth127 can review first thing monday and we'll publish a 2.3.1 binary that you can use and you're in business.

timwaterhouse:
No rush from me, thanks @dpastoor. I can wait until next week.

parsers/nmparser/diagonal_elements.go
```diff
@@ -10,108 +10,312 @@ import "fmt"
...
-	903:  42,
-	946:  43,
-	990:  44,
-	1035: 45,
-	1081: 46,
-	1128: 47,
-	1176: 48,
-	1225: 49,
-	1275: 50,
+	0:     0,
```
seth127, line 13: Do these need to be hard-coded like this? Is there a way to just create this map programmatically in the code? If not, I guess this is fine. 

Our upper limit is now 300 ETA's then? @timwaterhouse are we confident that this is more than we would ever realistically need?

parsers/nmparser/parse_lst_file_test.go
```diff
@@ -9,8 +9,8 @@ import (
 )
 
 func TestParseOBJV(t *testing.T) {
```
seth127, line 11: I'm fine with adding this test (assuming it passes) but these unit tests are largely broken and don't get run in CI, etc. for that reason.

The quasi-integration tests [here](https://github.com/metrumresearchgroup/babylontest) are what we really rely on. I would advocate adding another test model in [here](https://github.com/metrumresearchgroup/babylontest/tree/master/testdata/bbi_summary) and adding it to [the happy path list](https://github.com/metrumresearchgroup/babylontest/blob/master/bbi_summary_test.go#L28) in `bbi_summary_test.go`. Even if we don't actually write any new tests, that will at least cover this situation broadly.

@dpastoor @timwaterhouse if you can get me either a fake or real control stream for this (ideally with an output dir already populated, i.e. already run) then I can make the PR in babylontest for us.

parsers/nmparser/read_shk_test.go
```diff
@@ -55,42 +53,62 @@ func TestReadParseShkLines(t *testing.T) {
...
+				"TABLE NO.     1: First Order Conditional Estimation with Interaction: Problem=1 Subproblem=0 Superproblem1=0 Iteration1=0 Superproblem2=0 Iteration2=0",
+				"TABLE NO.     2: First Order Conditional Estimation with Interaction: Problem=1 Subproblem=0 Superproblem1=0 Iteration1=0 Superproblem2=0 Iteration2=0",
+				"TABLE NO.     3: First Order Conditional Estimation with Interaction: Problem=1 Subproblem=0 Superproblem1=0 Iteration1=0 Superproblem2=0 Iteration2=0",
+			},
 			context:   "3 methods",
 			etaCount:  4,
 			epsCount:  1,
 		},
+		{
+			lines: []string{
```
seth127, line 66: same point as above. These tests don't actually get run except periodically by a developer manually. Better to include in babylontest.

seth127 (changes requested):
looks good, but I wanted to add a test case to babylontest

parsers/nmparser/diagonal_elements.go
timwaterhouse: Never say never! I don't think I've ever had more than about 100 ETAs in a model, but that's not to say it can't be done.  300 is most likely more than enough in the interim, but @dpastoor [mentioned in the issue](https://github.com/metrumresearchgroup/babylon/issues/188#issuecomment-695207331) that this will be derived dynamically in the future.

seth127:
@dpastoor @timwaterhouse I changed this to be dynamic instead of just a bigger lookup table. Lemme know if this looks good with you guys:
https://github.com/metrumresearchgroup/babylon/pull/189/commits/96c9bc0eef2ea7f7bcbf13cc66274d0d369672ed

It passes all the tests (including Tim's new one) on this branch of babylontest:
https://github.com/metrumresearchgroup/babylontest/pull/21

dpastoor (commented):
Magnificent! Always happy to replace brute force with logic 🤩

timwaterhouse:
Looks good!

PR #191, opened 2020-10-07
Debugging issue Tim W found with empty parameter table

seth127:
We had an example furnished by @timwaterhouse of a model with 102 ETA's that parsed into an empty parameter table. 

The issue was that `ReadParamsAndOutputFromExt()` had an uncaught `bufio.ErrTooLong` because the lines in the `.ext` were longer than the buffer allowed and that was causing it to return an empty array that eventually got passed in to `ParametersData`. 

Increasing the buffer max solved the issue. His (very large) model had an average line size of ~75k. As far as I can see there is not technically a max for how big the lines could be, but I bumped it to 256k which should be plenty big.

### DIFF
--- a/parsers/nmparser/diagonal_elements.go
+++ b/parsers/nmparser/diagonal_elements.go
@@ -5,16 +5,42 @@ import (
 	"math"
 )
 
+// lowerDiagonalLengthToDimension takes l which is
+// the number of total elements in the lower diagonal
+//  matrix, and calculates the dimension (aka number of
+// diagonal elements) for the matrix.
+// Additionally, this function can be used to pass
+// a specific index from the array of lower diagonal
+// elements (as l). In that case, it will return one
+// of the following:
+// * If the index is a diagonal element:
+//   * The number of that diagonal element
+//   * true
+// * If the index is not a diagonal element:
+//   * 0
+//   * false
+// For example, ther are 6 elements in the lower
+// diagonal of a 3x3 matrix, the 1st, 3rd, and 6th
+// being diagonals. Calling this function on 1-6
+// would return the following:
+// lowerDiagonalLengthToDimension(1) => 1, true
+// lowerDiagonalLengthToDimension(2) => 0, false
+// lowerDiagonalLengthToDimension(3) => 2, true
+// lowerDiagonalLengthToDimension(4) => 0, false
+// lowerDiagonalLengthToDimension(5) => 0, false
+// lowerDiagonalLengthToDimension(6) => 3, true
 func lowerDiagonalLengthToDimension(l int) (int, bool) {
-
-	// estimate the dimension
+	​
+	// calculate the dimension
 	dim := int(math.Floor(math.Sqrt(2.0 * float64(l))))
-
+	​
 	// if not equal going back the other way then it's not a diagonal
 	check := (dim * (dim + 1)) / 2
-	is_diag := l == check
-
-	return dim, is_diag
+	if l != check {
+		return 0, false
+	}
+	​
+	return dim, true
 }
 
 func IndexAndIsDiag(i int) (int, bool) {
--- a/parsers/nmparser/diagonal_elements.go
+++ b/parsers/nmparser/diagonal_elements.go
@@ -19,7 +19,7 @@ import (
 // * If the index is not a diagonal element:
 //   * 0
 //   * false
-// For example, ther are 6 elements in the lower
+// For example, there are 6 elements in the lower
 // diagonal of a 3x3 matrix, the 1st, 3rd, and 6th
 // being diagonals. Calling this function on 1-6
 // would return the following:
@@ -30,16 +30,16 @@ import (
 // lowerDiagonalLengthToDimension(5) => 0, false
 // lowerDiagonalLengthToDimension(6) => 3, true
 func lowerDiagonalLengthToDimension(l int) (int, bool) {
-	​
+
 	// calculate the dimension
 	dim := int(math.Floor(math.Sqrt(2.0 * float64(l))))
-	​
+
 	// if not equal going back the other way then it's not a diagonal
 	check := (dim * (dim + 1)) / 2
 	if l != check {
 		return 0, false
 	}
-	​
+
 	return dim, true
 }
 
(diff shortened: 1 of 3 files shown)


parsers/nmparser/diagonal_elements.go
```diff
@@ -5,16 +5,42 @@ import (
...
+// For example, there are 6 elements in the lower
+// diagonal of a 3x3 matrix, the 1st, 3rd, and 6th
+// being diagonals. Calling this function on 1-6
+// would return the following:
+// lowerDiagonalLengthToDimension(1) => 1, true
+// lowerDiagonalLengthToDimension(2) => 0, false
+// lowerDiagonalLengthToDimension(3) => 2, true
+// lowerDiagonalLengthToDimension(4) => 0, false
+// lowerDiagonalLengthToDimension(5) => 0, false
+// lowerDiagonalLengthToDimension(6) => 3, true
```
seth127, line 31: this is unrelated to this issue; it's just a final tweak (and some docs) that I was trying to get into #189 but @dpastoor merged it before I could. So quick.

parsers/nmparser/utils.go
```diff
@@ -20,25 +23,29 @@ func createDiagonalBlock(num int) []int {
 // Checks the final estimation method against a list of non-gradient based methods.
 // Primarily used for deciding whether to look for a .grd file to parse
 func CheckIfNotGradientBased(results SummaryOutput) bool {
-	isNotGradientBased := funk.Contains([]string{
-		"MCMC Bayesian Analysis",
+		finalMethod := results.RunDetails.EstimationMethods[len(results.RunDetails.EstimationMethods)-1]
```
seth127, line 26: This is unrelated too, but @timwaterhouse 's model had
```
Estimation Method(s):
 - Stochastic Approximation Expectation-Maximization (No Prior)
 - Objective Function Evaluation by Importance Sampling (No Prior)
```
neither of which were triggering `isNotGradientBased` previously. This seemed like a better solution going forward.

utils/io.go
```diff
@@ -20,6 +20,11 @@ func ReadParamsAndOutputFromExt(path string) ([]string, error) {
 	}
 	defer inFile.Close()
 	scanner := bufio.NewScanner(inFile)
+
```
seth127, line 23: This is the real fix.

parsers/nmparser/utils.go
dpastoor: what is the concept here? This is now changing it if any of the methods are non-gradient based rather than the last one only right?

dpastoor (commented):
quick round of comments

utils/io.go
```diff
@@ -36,6 +41,16 @@ func ReadParamsAndOutputFromExt(path string) ([]string, error) {
...
 	}
 
+	// explicitly handle line that's too long
+	if scanner.Err() == bufio.ErrTooLong {
+		err_msg := fmt.Sprintf(
+			"ReadParamsAndOutputFromExt() failed on %s with `%s`",
+			path,
+			scanner.Err(),
+		)
+		panic(err_msg)
```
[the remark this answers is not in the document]
seth127: That makes sense. What do we think _should_ happen up the stack? This error will mean the 
`extData` (basically the array of Parma estimates) comes back empty. Do we want the summary to fail in that case, or continue to parse and return a summary with no param estimates? In my view, this is kind of a technical issue where it should only happen if we have something configured wrong. And it that case an error (with better formatting and further up the stack perhaps) is the right move. Thoughts?

parsers/nmparser/utils.go
```diff
@@ -20,25 +23,29 @@ func createDiagonalBlock(num int) []int {
...
-	}, results.RunDetails.EstimationMethods[len(results.RunDetails.EstimationMethods)-1])
+	}
+
+	checkMethods := make([]bool, 0)
+	for _, m := range notGradientMethods {
+		checkMethods = append(checkMethods, strings.Contains(finalMethod, m))
+	}
 
-	return isNotGradientBased
+	return utils.AnyTrue(checkMethods)
```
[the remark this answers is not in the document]
seth127: Good point. I'll make that change.

parsers/nmparser/utils.go
seth127: No it's that previously we were doing exact string matching which meant that "Importance Sampling" was included but "Importance Sampling (No Prior)" didn't match. Now we're doing substring matching so we don't have to check as many things and also because NONMEM seems to have a lot of different variations on what it prints out here.

seth127: To be clear, we are still only checking the final method. That is correct, right?

utils/io.go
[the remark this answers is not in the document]
dpastoor: yep - this is something that historically was something I personally did a poor job with (doing things like log.Fatal/panic/etc at the point of failure). For the purposes of the babylon CLI in general I would expect it to bubble all the way to the top then be a fatal error at that point (so exit code non zero + error returned). However, if other tooling was to say wrap this logic elsewhere it could separately process and pass that information to the user.

utils/io.go
```diff
@@ -36,6 +41,12 @@ func ReadParamsAndOutputFromExt(path string) ([]string, error) {
 		}
 	}
 
+	// explicitly handle line that's too long
+	if errors.Is(scanner.Err(), bufio.ErrTooLong) {
```
seth127, line 45: @dpastoor made this change you requested (and the small change in `notGradientMethods` above). Ready for you to take another (final?) look.

This passes it back out and it's eventually surfaced on line 62 or 106 of `cmd/summary.go`. I commented out lines 25 and 26 (the fix) above and ran the problem model and got:
```
$ bbi nonmem summary /Users/SethG/Downloads/70/70.lst 

FATA[0000] Attempting to parse /Users/SethG/Downloads/70/70.ext with ReadParamsAndOutputFromExt(): bufio.Scanner: token too long 

Process finished with exit code 1
```
which, you're right, is much better than panicing.

dpastoor: cool looks good

dpastoor (approved):
LGTM

PR #196, opened 2020-10-29
Release/2.3.1

seth127:
# Release notes

## Bug fixes

* Increased buffer size for reading lines from the `.ext` file to 256k. (#191)

* Changed checks for Bayesian and Non-gradient estimation methods to partial string matching so that more estimation methods are correctly matched. (#191)

* Changed lower-diagonal lookup to a dynamic function instead of a lookup table. This effectively means there is no longer a limit on the number of `ETA`'s that can be parsed from a model. (#189)

* Fixed bug where model files with a period in the name (i.e. `1000.1.ctl`) were being parsed incorrectly. (#192)


dpastoor (approved):
great work

PR #200, opened 2020-12-18
cleanup and dep upgrade

dpastoor:
This drops the server implementation used in the initial v1 that is likely not even functional at this point + does a full dep upgrade


seth127:
Closing and deleting per comment here https://github.com/metrumresearchgroup/bbi/pull/206#issuecomment-763276847

PR #202, opened 2020-12-30
WIP: params function

dpastoor:
Closes https://github.com/metrumresearchgroup/bbi/issues/219

this will offer a new command `bbi nonmem params` that will allow fast access to parameters only, especially in batch.

Prelim tests find compared to mapping over read_model %>% model_summary %>% param_estimates() this is ~ 1000x faster.

This becomes a big deal with bootstrap --> SSE where the time difference is, for example, ~4 minutes vs ~ 0.5 seconds to the final data frame

Prelim design doc available at: https://metworx.atlassian.net/wiki/spaces/DD/pages/56658449/rbabylon+fast+parameter+summaries


cmd/params.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+func params(cmd *cobra.Command, args []string) {
+	if debug {
+		viper.Debug()
+	}
+	var modelDirs []string
+	if dir != "" {
+		fi, err := ioutil.ReadDir(dir)
```
dpastoor, line 65:
```suggestion
        exdir, _ := homedir.Expand(dir)
		fi, err := ioutil.ReadDir(exdir)
```

PR #205, opened 2021-01-18
nm75 and updates

dpastoor:
This PR does 3 things for brevity:

* drops all server references from original bbi v1 which had a http api
* updates to handle nm75 shrinkage field
* updates the deps and checks in update from tidied go mod


seth127 (approved):
LGTM

PR #206, opened 2021-01-18
fmt: go fmt project

dpastoor:
go fmt project and cleaned up some misc top level files that were legacy

PR #208, opened 2021-01-25
Release 3.0.1

seth127:
# Release Notes

Primarily a patch to fix a bug when parsing output from NONMEM 7.5 runs (#205). Also ran `gofmt` and cleaned up some old files and documentation that had gotten stale and was no longer in use or accurate (#206).


seth127 (commented):
We tested the built binary from this release manually with the new `bbr` and everything looks good. Nothing changed on the release branch so we're going ahead with the release.

PR #215, opened 2021-04-19
fix/ eigenvalue parsing bug

### LINKED ISSUE #214: Condition number parsed incorrectly when eigenvalues fail to parse
Condition number is erroneously parsing as `1.0` in cases where NONMEM had written an extra line in between the heading and the eigenvalues.

Triggered in `parse_lst_file.go`
```go
ratio := 1.0 // If only 1 eigenvalue, the Condition Number is 1.0
if len(eigenvalues) >= 2 {
    ...
}
```
because the blank new line was incorrectly being parsed into `eigenvalues`, in which `len(eigenvalues)` was 1 and the `if` where we calculated the condition number was skipped.

### LINKED ISSUE #216: Eigenvalue issues heuristic
If any of the of the eigenvalues `<= 0`, or if NONMEM needs to force them to be positive, this heuristic should be triggered. The second condition can be identified by looking for `Forcing positive definiteness` in the `.lst` file.

seth127:
Closes #214 
Closes #216 

Also refactors to pull condition number from `-1000000003` line of `.ext`, when present, instead of `.lst` file.

### DIFF
--- a/parsers/nmparser/parse_lst_file.go
+++ b/parsers/nmparser/parse_lst_file.go
@@ -310,6 +310,8 @@ func ParseLstEstimationFile(lines []string) SummaryOutput {
 			runHeuristics.ParameterNearBoundary = true
 		case strings.Contains(line, "COVARIANCE STEP ABORTED"):
 			runHeuristics.CovarianceStepAborted = true
+		case strings.Contains(line, "Forcing positive definiteness"):
+			runHeuristics.EigenvalueIssues = true
 		case strings.Contains(line, "EIGENVALUES OF COR MATRIX OF ESTIMATE"):
 			allCondDetails = parseConditionNumberLst(lines, i, allCondDetails)
 		default:
@@ -319,6 +321,11 @@ func ParseLstEstimationFile(lines []string) SummaryOutput {
 
 	runHeuristics.HasFinalZeroGradient = parseGradient(gradientLines)
 	runHeuristics.LargeConditionNumber = getLargeConditionNumber(allCondDetails)
+	for _, cd := range allCondDetails {
+		if (cd.ConditionNumber <= 0) && (cd.ConditionNumber != DefaultFloat64){
+			runHeuristics.EigenvalueIssues = true
+		}
+	}
 
 	var finalParameterEst ParametersResult
 	var finalParameterStdErr ParametersResult
@@ -457,7 +464,10 @@ func calculateConditionNumber(lines []string, start int) float64 {
 	}
 
 	sort.Float64s(eigenvalues)
-	ratio := eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
+	ratio := -1.0 // if eigenvalues contain a zero we return condition number of -1
+	if eigenvalues[0] != 0 {
+		ratio = eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
+	}
 	return ratio
 }
 
--- a/parsers/nmparser/parse_lst_file.go
+++ b/parsers/nmparser/parse_lst_file.go
@@ -406,12 +406,12 @@ func parseConditionNumberLst(lines []string, start int, allCondDetails []Conditi
 	// always modify the most recently created ConditionNumDetails
 	condDetails := &allCondDetails[len(allCondDetails)-1]
 
-	condDetails.ConditionNumber = calculateConditionNumber(lines, start)
+	condDetails.ConditionNumber = mustCalculateConditionNumber(lines, start)
 
 	return allCondDetails
 }
 
-func calculateConditionNumber(lines []string, start int) float64 {
+func mustCalculateConditionNumber(lines []string, start int) float64 {
 
 	// go until line of ints
 	for i, line := range lines[start:] {
--- a/parsers/nmparser/parse_lst_file.go
+++ b/parsers/nmparser/parse_lst_file.go
@@ -2,6 +2,7 @@ package parser
 
 import (
 	"bbi/utils"
+	"fmt"
 	"math"
 	"sort"
 	"strconv"
@@ -457,7 +458,7 @@ func calculateConditionNumber(lines []string, start int) float64 {
 		for _, s := range strings.Fields(line) {
 			eigenvalue, err := strconv.ParseFloat(s, 64)
 			if err != nil {
-				panic("Attempting to calculate condition number but could not parse eigenvalues")
+				panic(fmt.Sprintf("Attempting to calculate condition number but could not parse eigenvalues -- %v", err))
 			}
 			eigenvalues = append(eigenvalues, eigenvalue)
 		}
--- a/parsers/nmparser/parse_lst_file.go
+++ b/parsers/nmparser/parse_lst_file.go
@@ -440,28 +440,30 @@ func calculateConditionNumber(lines []string, start int) float64 {
 		}
 	}
 
-	// go until another blank line and build eigenvalues vector
+	// go until non-blank line and build eigenvalues vector
 	var eigenvalues []float64
-	for i, line := range lines[start:] {
-		for _, s := range strings.Fields(line) {
-			eigenvalue, err := strconv.ParseFloat(s, 64)
-			if err == nil {
-				eigenvalues = append(eigenvalues, eigenvalue)
+	eigenstarted := false
+	for _, line := range lines[start:] {
+		sub := strings.TrimSpace(line)
+		if len(sub) == 0 {
+			if eigenstarted {
+				break
+			} else {
+				continue
 			}
 		}
 
-		sub := strings.TrimSpace(line)
-		if len(sub) == 0 {
-			start = start + i + 1
-			break
+		eigenstarted = true
+		for _, s := range strings.Fields(line) {
+			eigenvalue, err := strconv.ParseFloat(s, 64)
+			if err != nil {
+				panic("Attempting to calculate condition number but could not parse eigenvalues")
+			}
+			eigenvalues = append(eigenvalues, eigenvalue)
 		}
 	}
 
-	ratio := 1.0 // If only 1 eigenvalue, the Condition Number is 1.0
-	if len(eigenvalues) >= 2 {
-		sort.Float64s(eigenvalues)
-		ratio = eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
-	}
-
+	sort.Float64s(eigenvalues)
+	ratio := eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
 	return ratio
 }
--- a/parsers/nmparser/parse_lst_file.go
+++ b/parsers/nmparser/parse_lst_file.go
@@ -280,7 +280,7 @@ func ParseLstEstimationFile(lines []string) SummaryOutput {
 			// starting new estimation method, make new details objects
 			method := strings.TrimSpace(strings.Replace(line, "#METH:", "", -1))
 			allOfvDetails = append(allOfvDetails, NewOfvDetails(method))
-			allCondDetails = append(allCondDetails, NewConditionNumDetails(method))
+			allCondDetails = append(allCondDetails, NewConditionNumDetails(method, DefaultFloat64))
 		case strings.Contains(line, "#OBJV"):
 			allOfvDetails = parseOFV(line, allOfvDetails)
 		case strings.Contains(line, "CONSTANT TO OBJECTIVE FUNCTION"):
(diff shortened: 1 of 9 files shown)


parsers/nmparser/parse_lst_file.go
```diff
@@ -440,28 +441,30 @@ func calculateConditionNumber(lines []string, start int) float64 {
...
 
-	ratio := 1.0 // If only 1 eigenvalue, the Condition Number is 1.0
-	if len(eigenvalues) >= 2 {
-		sort.Float64s(eigenvalues)
-		ratio = eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
+		eigenstarted = true
+		for _, s := range strings.Fields(line) {
+			eigenvalue, err := strconv.ParseFloat(s, 64)
+			if err != nil {
+				panic(fmt.Sprintf("Attempting to calculate condition number but could not parse eigenvalues -- %v", err))
```
dpastoor, line 461: generally, we never want to panic unless there is a really bad reason to. Unfortunately, throughout this codebase, and some of pkgr, this was not always well held as panic is an easy escape hatch of just 💣 .

If a function is going to panic on such a situation, it should communicate it in the function name, generally using the term `must` , so like `mustParse`

If not, instead can have the function signature return the value + error, so then the if err != nil just return right there like 0, err

The idea becomes then that error can bubble all the way up the stack to the final output, or be intercepted and handled otherwise, but that is something for some other place in the code to decide what to do about it.

parsers/nmparser/parse_lst_file.go
```diff
@@ -440,28 +441,30 @@ func calculateConditionNumber(lines []string, start int) float64 {
...
+			}
+			eigenvalues = append(eigenvalues, eigenvalue)
+		}
 	}
 
+	sort.Float64s(eigenvalues)
+	ratio := eigenvalues[len(eigenvalues)-1] / eigenvalues[0]
```
dpastoor, line 468: could the code get here if no eigenvalues were found - if so len(0) - 1 could cause problems. 

Also, could there be an eigenvalue of 0, in which case divide by 0

dpastoor (changes requested):
let me know if you need me to explain more the error handling, this will likely require adjustments in a couple more upstream places as well

parsers/nmparser/parse_lst_file.go
seth127: I actually started to code this up and then realized that nothing for about three levels above this expects an error to come through so I just got lazy and made it panic. My thinking was that this should never be a user-facing error. In other words, it should only trigger if we parse something incorrectly, in which case spewing the ugly stack trace is probably as good as anything because we'll just want them to come to us with it.

Alternatively, we could handle it a few levels up and just log the error and return something like `nil` for the condition number. But that didn't feel right to me, honestly. Your thoughts?

parsers/nmparser/parse_lst_file.go
seth127: > could the code get here if no eigenvalues were found - if so len(0) - 1 could cause problems.

I think no. I had an explicit check for that in here (that also panicked) but it felt excessive. I think the only way we would get here with length==0 is if the `EIGENVALUES` header was the _last thing_ in the file. Because if there was a number afterwards it would parse that as an eigenvalue, and if there was anything else it would panic on line 461.

> could there be an eigenvalue of 0, in which case divide by 0
hmmm, I hadn't considered this. We never checked for this before but we could. In that case... what should we do? Seems like we either have to error or substitute some very tiny number instead. But the former is what would already happen and the latter feels very wrong to me because that would just return some very large (and probably wrong) condition number.

Honestly this math is a little over my head, but a little googling tells me that a) [it is _possible_ to have an eigenvalue equal to zero](https://math.stackexchange.com/a/1050849) and b) that [might mean](https://en.wikipedia.org/wiki/Condition_number#Nonlinear) the condition number is technically infinite. Any guidance on this scenario is appreciated.

dpastoor: i'd defer to @kylebaron or @timwaterhouse

parsers/nmparser/parse_lst_file.go
dpastoor: yeah i guessed as much, and could also see just using the mustParse() nomenclature and leaving the panic. I would also pass the specific value that failed on parsing as well (s) to help with diagnostics.

parsers/nmparser/parse_lst_file.go
timwaterhouse: Yeah, it should be possible (but unlikely) to get eigenvalues of zero if you don't force the R matrix to be positive definite (all eigenvalues > 0: default for non-EM methods is to not force positive definiteness). I think the condition number doesn't mean too much if the matrix has zero or negative eigenvalues, so maybe return a very large number (or infinity if possible). Even if the lowest eigenvalue is a very small positive number, that still indicates a poor fit so a large condition number is appropriate. @kylebaron?

kylebaron: I'm thinking that you just won't get anything if it can't come up with positive eigenvalues (like it'll say "COVARIANCE MATRIX UNOBTAINABLE" or something like that or it just won't give you anything).

I think NONMEM is making all these decisions too and implementing it in the `.ext` file.  If we take that output (or lack of output) bbi won't be in conflict with what NONMEM is doing and non assumptions needed on how to parse etc.

kylebaron: No eigenvalues reported

```
0S MATRIX ALGORITHMICALLY SINGULAR
0COVARIANCE MATRIX UNOBTAINABLE
```

kylebaron: All the eigenvalues are positive
```
 WARNING: R MATRIX ADJUSTED TO BE MADE POSITIVE DEFINITE.
 INTERPRET VARIANCE-COVARIANCE OF ESTIMATES WITH CARE.
```

kylebaron: All the eigenvalues are positive

```
 Number of Negative Eigenvalues in Matrix=          18
 Most negative value=  -2464456983.1781917
 Most positive value=   24673139456761.383
 Forcing positive definiteness
 Root mean square deviation of matrix from original=    2.1141151552555096E-004
```

seth127: Ok, soooo... some of that went a little over my head, but let me try to sum up. I can see two easy ways to code this. @timwaterhouse @kylebaron let me know which you'd prefer:

1. If the _smallest_ eigenvalue is 0 or negative, substitute in `0.0000001`. This should make the returned condition number very large, and would trigger the associated heuristic.
2. If the _smallest_ eigenvalue is 0 or negative, return something like `NULL` for the condition number. In this case, should we trigger the `large_condition_number` heuristic?

There's also the option Kyle mentioned of trying to parse it from line `-1000000003` of the `.ext` file. I'd prefer _not_ to do this just because `bbi` basically tries to get as much as possible from the `.lst` file, so that it could conceivably parse _only_ an `.lst` file without anything else. Maybe this is a silly thing to try to maintain... I don't know. But at least at this point I feel like I want to maintain it unless there's a good reason not to.

Also, if we go with parsing from the `.ext` there are still the two questions from point 2 above, in cases where that line is _not_ present in the `.ext`

> return something like `NULL` for the condition number. In this case, should we trigger the `large_condition_number` heuristic?

kylebaron: @seth127 You also have to think about precision that is reported; if you ratio the eigenvalues you find in the lst then you will always get a different condition number than NONMEM is saying; 

## lst

When this is in the lst
```
2.87E-03  3.14E-02  4.69E-02  1.11E-01  1.63E-01  2.68E-01  
5.31E-01  6.61E-01  1.09E+00  1.34E+00  1.35E+00  1.73E+00
1.88E+00  2.18E+00  3.61E+00
```
you calculate
```
         3.61E+00 / 2.87E-03   = 1257.84
```

## ext
When this is in the ext file

```
 -1000000003  1.25555E+03  2.87348E-03  3.60780E+00
```

you calculate

```
3.60780E+00 / 2.87348E-03   = 1255.551
```

And nonmem thinks the answer is: `1255.55`

It's not the end of the world to be off by 2.3; but like the parameter estimates, I think the ext file is the place to go for quantitative results.

seth127: That makes sense. It looks like what we do with the parameter estimates is parse them from the `.lst` and then, if there's an `.ext` file present, overwrite them with what we parse from the `.ext`. 

We could do the same thing with condition number. In that case, I would do:

* If parsing from `.ext` and `-1000000003` line is _not_ present, return `NULL` for condition number
* If parsing from `.lst` and smallest eigenvalue is <= 0 , return `NULL` for condition number

In both cases, should we trigger the `large_condition_number` heuristic flag or not?

timwaterhouse: I agree with getting these from the `.ext` file if possible.  As for what to return for a heuristic flag, I'm trying to test what NONMEM does with negative eigenvalues if it doesn't force positive definiteness, but it turns out that NONMEM *really* wants to force positive definiteness, even when setting the $COV options which should apparently stop that from happening. It will report that there *were* negative eigenvalues (like [this](https://github.com/metrumresearchgroup/bbi/pull/215#discussion_r616693552)), but then do this correction and report all positive eigenvalues. @kylebaron do you happen to know if there's a way around that?

What currently happens if bbi doesn't detect eigenvalues in the `.lst`? The user needs to specify `PRINT=E` in the control stream to get those output anywhere, so some kind of heuristic for missing eigenvalues would be helpful as a reminder. The same might be good for that line missing from the `.ext`. Possibly a separate heuristic for non-positive eigenvalues but as I said I'm finding it really hard to get NONMEM to actually output negative values since it always wants to force positive definiteness, so that might never come up.

kylebaron: Good questions, @timwaterhouse ; I haven't looked into any options but did have the feeling like (1) either the eigenvalues were positive [great!] or (2) they were forced to be positive or (3) you don't get anything.  And having a heuristic for (2) [the eigenvalues are positive at the end of the day but they didn't start out that way] would be really useful to know for a run. And it seems like the only way to get that now is by scrolling the `.lst` file. 

Looking at this option:
```
 FPOSDEF=n (NM74)
      By default FPOSDEF(forced positive  definite)=0.   If  FPOSDEF=1,
      then  if  the Rmatrix is not positive definite, it will be forced
      positive definite.  If PRECOND>0, this will occur after the  PRE-
      CONDth try.
```
Seems like (by default) it shouldn't force?

And I think we were getting this on the EM runs since it defaults to 3 then:

```
 POSDEF=n (NM75)
      If POSDEF=1, then if the R matrix is not  positive  definite,  it
      will  be  forced positive definite, by method of shifting the ei-
      genvalues by an amount slightly greater than  the  most  negative
      eigenvalue.  If POSDEF=2, then R matrix is made positive definite
      by making the negative valued eigenvalues approximately 1/100  of
      the least positive eigenvalue. **If POSDEF=3**, then R matrix is made
      positive definite by using the absolute values of the eigenvalues
      POSDEF  is  0 (no correction) by default for classical covariance
      step, and **3 for EM methods**.
      The preconditioning method is a more sophisticated, but time-con-
      suming  method of making the R matrix positive definite (See PRE-
      COND).
      Default is 0.
```

timwaterhouse: @kylebaron  Yeah, I've tried `FPOSDEF=0` and `POSDEF=0` but it keeps forcing it anyway. Tried on a METHOD=1 model as well, but it won't take no for an answer.

kylebaron: Oh interesting.  I guess i never paid close attention to it before.  But at least having that heuristic would be good. Would have helped on most-recent project.

seth127: Ok, so the proposal is to add another heuristic, which I'm not quite sure what to call (`eigenvalue_issues`?) that would trigger if (any of):

* smallest eigenvalue is <= 0
* eigenvalue line (`-10000003`) is missing from `.ext`
* any eigenvalues _were_ negative, but potentially got forced back to positive (look for "`Negative Eigenvalues in Matrix`" in the `.lst`?)

I realize these are all slightly different (or maybe overlapping?) but I don't think we want to have too many heuristics. The docs for `?model_summary` would explain that this flag can mean a few different things and "the user should check...".

How does that all sound? If we like that, then we'll continue on to discussing what to return for the "`NULL`" value... (it's a bit of a tangent, so I want to get this straight first)

kylebaron: Just kicking around some ideas (I know this is going to seem like a lot, but getting some ideas out):

- Agree we don't want too many heuristics / red flags coming up
- I feel like we need to know if the user both requested the covariance matrix and requested eigenvalues get printed (it seems like if you don't request them, they don't appear in either the `.lst` or the `.ext` files; @timwaterhouse can you confirm this?)
- If the user didn't request it we shouldn't warn that there was an issue with eigenvalues
- If the user did request eigenvalues and the covariance matrix was unobtainable or the cov step fails, maybe warn about the covariance matrix only
- If the user requested eigenvalues and NONMEM had to force positive definiteness, then issue an eigenvalue warning
  - You could look for something in that text; I'd tend toward `Forcing positive definiteness` but `Negative Eigenvalues` could work too
- If the user requested eigenvalues and NONMEM reports a negative eigenvalue, then issue an eigenvalue warning (honestly I doubt this will happen but maybe it could)

timwaterhouse: Right, if you don't use `$COV ... PRINT=E` then there are no eigenvalues in either `.lst` or `.ext` files. Agree that we shouldn't warn about missing eigenvalues if they're not requested. This should be flagged by `EIGENVLS. PRINTED:              NO` in the `.lst` file.

@seth127, I'm good with your proposed `eigenvalue_issues` (or whatever) instead of a bunch of separate ones.

dpastoor: with respect to lst vs ext - I would likewise go straight to the ext. That ship has now definitively sailed. At one point, with an objective of being similar to PsN sumo, we thought to maybe be able to give back results with only lst. At this point, its more hastle than its worth and if anything i'd rather rely on lst less and take advantage of more structured data

seth127:
@dpastoor this is ready for review again. I tagged https://github.com/metrumresearchgroup/bbi/releases/tag/3.0.3-beta.1 for testing.

seth127:
@dpastoor this is ready for your re-review. @timwaterhouse tested it on his project and it worked, plus we have tests now.

Note: Drone was failing on this branch, which was [fixed as discussed here](https://github.com/metrumresearchgroup/bbi/issues/217#issuecomment-848192663) as part of a red herring. That is [still ongoing](https://github.com/metrumresearchgroup/bbi/issues/217#issuecomment-848201666), but I don't think it effects this PR, mostly because it's been going on for 6 months at least and the flaky test is totally unrelated.

dpastoor:
Lgtm!

PR #221, opened 2021-07-13
Release v3.0.3

seth127:
Merging v3.0.3 changes into `master`. All changes from https://github.com/metrumresearchgroup/bbi/pull/215

# Release Notes

* Fixed a bug where `condition_number` would come back as `1.0` in situations where it shouldn't. (#214)
* Added `eigenvalue_issues` heuristic, triggered when any of the of the eigenvalues <= 0, or if NONMEM needs to force them to be positive. (#216)


seth127:
Merging this without review because it only contains changes from a single PR (#215) which got plenty of review already.

Issue #227, opened 2021-08-30
Bad import of `number_of_data_records` when using $INFN?


dpolhamus:
I'm not sure exactly where the problem is here, but when generating a `model_summary` of a model in which I've used `$INFN` to subset the dataset (i.e., instead of `IGNORE` in `$DATA`) it seems like `nonmem_summary` sets `number_of_data_records` to a large negative number (-999999999).  

The lst file reports the correct number of records (after subsetting).

Here's example offending `$INFN`, in this case keeping records in which ANL01FL==1:

```
$INFN
  IF(PRED_IGNORE_DATA_TEST==1) THEN
    PRED_IGNORE_DATA=0
    IF(ANL01FL.NE.1) PRED_IGNORE_DATA=1
    RETURN;
  ENDIF
```

kyleam:
The -999999999 is coming from bbi:

https://github.com/metrumresearchgroup/bbi/blob/43f1495bf78e6a6f41043889c76d8068cfe70e4c/parsers/nmparser/structs.go#L184

It initializes `RunDetails` with this value.  `ParseRunDetails` will try to set `runDetails.NumberOfDataRecords` from the "NO. OF DATA RECS IN DATA SET" line in the lst file:

https://github.com/metrumresearchgroup/bbi/blob/43f1495bf78e6a6f41043889c76d8068cfe70e4c/parsers/nmparser/parse_run_details.go#L84-L85

Just as a test, if I drop that line from bbr's `inst/model/nonmem/basic/1/1.lst`, I see the placeholder value:

```
$ bbi nonmem summary --json inst/model/nonmem/basic/1/1 | jq .run_details.number_of_data_records
799
# <drop "NO. OF DATA ..." line>
$ bbi nonmem summary --json inst/model/nonmem/basic/1/1 | jq .run_details.number_of_data_records
-999999999
```

> The lst file reports the correct number of records (after subsetting).

Hmm, so if it's there, then `ParseRunDetails` should be able to extract it. @dpolhamus, does the line in the lst file look slightly different when `$INFN` is used?

dpolhamus:
Here's an excerpt from an `lst` with a `$INFN` block:

```
0FORMAT FOR DATA:
 (22(3E21.0/))

 TOT. NO. OF DATA RECS:      1707
 TOT. NO. OF OBS RECS:      190
 TOT. NO. OF INDIVIDUALS:      205
0LENGTH OF THETA:   5
```

[...]

PR #228, opened 2021-09-01
ParseRunDetails: parse NumberOfDataRecords for $INFN output

kyleam:
In gh-227, @dpolhamus reported that the number of data records wasn't being set properly when `$INFN` is used.  The lst file doesn't have the "TOT. NO. OF DATA RECS:" line that `ParseRunDetails` looks for, but there is a "NO. OF DATA RECS IN DATA SET:" line.  The last commit of this series falls back to using that value, assuming there is only one "NO. OF DATA RECS IN DATA SET:" line and that its value is a reasonable value for `NumberOfDataRecords`.

The first two commits are related fixes (or at least what I think are fixes).  One of them gets the tests in `parsers/nmparser/parse_run_details_test.go` to a passing state.  @seth127, would you prefer that I drop these commits and add the test in metrumresearchgroup/bbitest instead?

(Also, I'm being a bit lazy with the new test and duplicating `RunDetails01Results` aside from one item.  I can rework that...)

```
$ (cd parsers/nmparser && go test -v -run TestParseRunDetails)
=== RUN   TestParseRunDetails
--- PASS: TestParseRunDetails (0.00s)
=== RUN   TestParseRunDetailsInfn
--- PASS: TestParseRunDetailsInfn (0.00s)
PASS
ok  	bbi/parsers/nmparser	0.195s
```

---

And here some of my manual tests with bbr's `inst/model/nonmem/basic/1/1.lst`.

```sh
# On bbr's 1.1.2.7001-7-g1802d63, no changes to lst file.
$ grep 'DATA RECS' inst/model/nonmem/basic/1/1.lst
 NO. OF DATA RECS IN DATA SET:      799
$ bbi nonmem summary --json inst/model/nonmem/basic/1/1 | jq .run_details.number_of_data_records

# Change lst file to match gh-227 case: "TOT. NO. OF DATA RECS"
# but no "NO. OF DATA RECS IN DATA SET"
$ grep 'DATA RECS' inst/model/nonmem/basic/1/1.lst
 TOT. NO. OF DATA RECS:      1707
$ bbi nonmem summary --json inst/model/nonmem/basic/1/1 | jq .run_details.number_of_data_records
1707
```

### DIFF
--- a/parsers/nmparser/parse_run_details.go
+++ b/parsers/nmparser/parse_run_details.go
@@ -83,6 +83,9 @@ func ParseRunDetails(lines []string) RunDetails {
 			runDetails.NumberOfObs, _ = strconv.ParseInt(replaceTrim(line, "TOT. NO. OF OBS RECS:"), 10, 64)
 		case strings.Contains(line, "NO. OF DATA RECS IN DATA SET:"):
 			runDetails.NumberOfDataRecords, _ = strconv.ParseInt(replaceTrim(line, "NO. OF DATA RECS IN DATA SET:"), 10, 64)
+		// When using $INFN, the above line isn't present (see gh-227).
+		case strings.Contains(line, "TOT. NO. OF DATA RECS:") && runDetails.NumberOfDataRecords == DefaultInt64:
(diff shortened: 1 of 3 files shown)


seth127 (approved):
Looks good for the fix and the tests. Our CI is currently failing, which I was looking into a bit yesterday, but that's not related to this so I think it's fine to merge this to develop.

PR #230, opened 2021-09-07
Lint misc

elisarver:
Hey all, this is a series of commits to conform to our common golangci-lint configuration. This gets a number of todos on the board, and cleans up a lot of stuff that could cause errors in weird edge cases.


cmd/summary.go
```diff
@@ -101,7 +105,7 @@ func summary(cmd *cobra.Command, args []string) {
 	var modelResults jsonResults
 
 	for w := 1; w <= workers; w++ {
-		go func(w int, modIndex <-chan int, results chan<- modelResult) {
+		go func(_ int, modIndex <-chan int, results chan<- modelResult) {
```
dpastoor, line 108: could probably just drop this instead - since w isn't used anywhere and just simplify the func to modIndex,results.

dpastoor (commented):
for the linting with no uses - i'm guessing but just want to confirm that it doesn't like only check within that file, but rather across the entire package that the global var isn't used?

elisarver:
> for the linting with no uses - i'm guessing but just want to confirm that it doesn't like only check within that file, but rather across the entire package that the global var isn't used?

So used in go terms is looking for *reads* of a value. You can set things forever until you're blue in the face, but if you never *read* it, it's "unused." Each of these tools has the entire AST of all packages in memory, so they have a complete view of the program. The IDE redundantly does the same check, and each of these was grey, meaning its model also found them unused. (which is based on the gopls service)

cmd/summary.go
elisarver: Yeah, a lot of them can be dropped, so long as we didn't unintentionally not use them. Hence the TODO

cmd/nonmem.go
```diff
@@ -137,122 +133,120 @@ var parallelRegexesToRemove []string = []string{
...
 
 func init() {
 	RootCmd.AddCommand(nonmemCmd)
 
-	//NM Selector
+	// NM Selector
 	const nmVersionIdentifier string = "nm_version"
 	nonmemCmd.PersistentFlags().String(nmVersionIdentifier, "", "Version of nonmem from the configuration list to use")
-	viper.BindPFlag(nmVersionIdentifier, nonmemCmd.PersistentFlags().Lookup(nmVersionIdentifier))
+	errpanic(viper.BindPFlag(nmVersionIdentifier, nonmemCmd.PersistentFlags().Lookup(nmVersionIdentifier)))
```
shairozan, line 186: I really had to think through the use of this in the execution goroutines a lot. For local execution it would seem to be fine because you want to interrupt everything if something fails, but in SGE execution I had to step back and think about it again. The initial phase that the user interacts with just basically puts a script onto the SGE nodes that runs locally there, so if failures occur there, it's all written out into the out file. So even if *something* failed and we panic'd in the goroutines for execution, it would still only affect that run. 

What it would do is interfere with the post execution scripts, which I don't know if is really a problem since I don't know of anyone using them. The panic would basically keep the orchestrator from getting to the point of running the script and indicating a status of failure

elisarver: Well if a panic is bad, what about a silent failure?

shairozan:
A silent failure would still be fine. What would happen is any error state
would be wrapped up in the status of the job by the manager and collected
into the post execution script

cmd/nonmem.go
elisarver: > A silent failure would still be fine. What would happen is any error state would be wrapped up in the status of the job by the manager and collected into the post execution script

I think this may be rendered moot as all of these panics would happen on create of the rootcmd and its dependent children. In other words, the application will **always** panic on startup or **never** panic on startup.

cmd/local.go
```diff
@@ -406,9 +406,9 @@ func local(cmd *cobra.Command, args []string) {
 	}
 }
 
-//WriteGitIgnoreFile takes a provided path and does best attempt work to write a "Exclude all" gitignore file in the location
+//WriteGitIgnoreFile takes a provided path and does best attempt work to write a "Exclude all" gitignore file in the location.
 func WriteGitIgnoreFile(filepath string) {
-	utils.WriteLines([]string{"*"}, path.Join(filepath, ".gitignore"))
+	errpanic(utils.WriteLines([]string{"*"}, path.Join(filepath, ".gitignore")))
```
shairozan, line 411: This is a scenario where it would not be moot since WriteGitIgnore is in the execution path. It's part of the setup phase for making sure random nonmem files don't get added to git in the output structures. These are the pathways we have to worry about the panics in. 

The other ones mentioned are in fact during startup, but using it all over the place could have the unexpected consequence of killing the ability to fire off postexecution scripts. Their main purpose was to notify a user whether a job succeeded or failed

elisarver: So you have a situation where:

1. a function touches the filesystem
2. the functions in the filesystem can fail, leaving everything in a weird state
3. the function does not return error
4. you don't want to throw away your data

what do we do in these situations?

shairozan: In a scenario where the function doesn't return an error, it needs to get surfaced to the execution phases (the turnstile interface commands). If the error gets surfaced there, the manager will at least log to its std out that an error occurred. 

In an SGE environment the entire output of the bbi run is stored in a log file. Secondarily if the job was not successful (ie collected an error) if a user is utilizing the post-execution script hooks, it would indicate whether or not that run was successful. The goal would just be to elevate it to the manager so that a ConcurrentError can be recorded.

shairozan (approved):
At least we won't have to worry about the linter going forward :)

PR #241, opened 2021-09-17
Wrapt

elisarver:
Update tests around the use of wrapt. Some tests were turned into proper tests over loops of data where available and where the core test was identical. This results in a more hierarchical structure for the test, and eliminates as much global state as possible.


.drone.yml
```diff
@@ -28,7 +28,7 @@ steps:
       - mkdir -p /tmp/${DRONE_BUILD_NUMBER} ; git rev-parse HEAD > /tmp/${DRONE_BUILD_NUMBER}/bbitest_hash
       - if [ ! -z $BBITEST_BRANCH ]; then git checkout $BBITEST_BRANCH; fi
     environment:
-      BBITEST_BRANCH: master
+      BBITEST_BRANCH: wrapt
```
shairozan, line 31: I just approved the bbitest repo, what might make more sense is to merge that first and then swap this back to master.

parsers/nmparser/parse_final_parameter_estimates_test.go
```diff
@@ -1,132 +1,132 @@
...
-}
+func TestParseFinalParameterEstimatesFromLst(tt *testing.T) {
+	tests := []struct {
+		name  string
+		input []string
+		want  ParametersResult
+	}{
+		{
+			name: "final parameter estimate",
+			input: []string{
```
shairozan, line 17: I feel like storing strings as multiline with '`'  would look cleaner. Probably not worth the effort if the functions are expecting []string though. Feels like functions that are refactorable for just taking readers / writers

shairozan: Not for right now obviously. Just something to think about as we start refactoring

runner/prepare_to_execute_test.go
```diff
@@ -1,84 +1,87 @@
...
+func TestPrepareForExecution(tt *testing.T) {
+	type pathtest struct {
+		name string
+		in   []string
+		out  []string
 	}
+
+	var tests []pathtest
+
 	if runtime.GOOS == "windows" {
```
shairozan, line 19: May make sense to make this build specific to windows or move to a `_windows_test.go` file later to keep things clean

shairozan (approved):
Passing all of the existing E2E tests and the conversions to wrapt look pretty straight forward since we were already using testify Assert / Require. Also glad to see moving a lot of the globals farther into the test definitions since I know that was a problem we had to work through on the execution side as well. 

Pointed out a few things we can think about as we refactor further, but nothing to keep this from going since this was primarily a refactor of tests to the wrapt design.

runner/prepare_to_execute_test.go
elisarver: I was thinking the same, but at the same time I don't have a windows machine to run it on.

.drone.yml
elisarver: yeah done

parsers/nmparser/parse_final_parameter_estimates_test.go
elisarver: i'd use []byte normally

PR #242, opened 2021-09-20
Merge bbitest

elisarver:
This also conforms to our linter in order to not fail that phase.


shairozan (approved):
Was a pain to get running, but at least we've removed the need for a separate project now. It took longer to start since the test files occurred at first, but once I got to the data directory, it went pretty quick

PR #247, opened 2021-09-23
Test/speedup

shairozan:
* Remove duplicitous scenarios from tests
* Update drone pipeline to fail faster


elisarver (approved):
This should lighten the load.

PR #251, opened 2021-10-01
Adding mrgvalidate style Test Id's

### LINKED ISSUE #246: Add mrgvalidate Test Id's to any tests that are missing them
As described in https://metrumresearchgroup.github.io/mrgvalprep/articles/basic_usage.html#test-ids

Some should have already been added from #239 and #233 but there will surely be some left that still need them.

This will also involve mapping these to the updated requirements and stories in googlesheets or whatever end up using for #238

seth127:
Closes #246

### DIFF
--- a/runner/consts_test.go
+++ b/runner/consts_test.go
@@ -2,28 +2,9 @@ package runner
 
 import (
 	"testing"
-
+	"github.com/metrumresearchgroup/bbi/utils"
 	"github.com/metrumresearchgroup/wrapt"
 )
-
-// TODO: remove initResults if we find no use
-/* func initResults() map[string]int {
-	var results01 = make(map[string]int)
-	results01["run001.clt"] = 1
-	results01["run001.coi"] = 1
-	results01["run001.cor"] = 1
-	results01["run001.cov"] = 1
-	results01["run001.cpu"] = 1
-	results01["run001.ext"] = 1
-	results01["run001.grd"] = 1
-	results01["run001.lst"] = 1
-	results01["run001.phi"] = 1
-	results01["run001.shk"] = 1
-	results01["run001.shm"] = 1
-	results01["run001.xml"] = 1
-
-	return results01
-} */
 
 func TestEstOutputFilesByRun(tt *testing.T) {
 	var tests = []struct {
@@ -48,8 +29,10 @@ func TestEstOutputFilesByRun(tt *testing.T) {
 			},
 		},
 	}
+
+	testId := "UNIT-RUN-001"
 	for _, test := range tests {
-		tt.Run(test.in, func(tt *testing.T) {
+		tt.Run(utils.AddTestId(test.in, testId), func(tt *testing.T) {
 			t := wrapt.WrapT(tt)
 
 			got := EstOutputFilesByRun(test.in)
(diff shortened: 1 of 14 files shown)


runner/consts_test.go
```diff
@@ -2,29 +2,10 @@ package runner
 
 import (
 	"testing"
-
+	"github.com/metrumresearchgroup/bbi/utils"
 	"github.com/metrumresearchgroup/wrapt"
 )
 
-// TODO: remove initResults if we find no use
```
seth127, line 9: @elisarver made this comment 3 weeks ago. I think we're fine removing this, unless someone has a reason to keep it. Everything passes without it though.

runner/estimate_model_test.go
```diff
@@ -7,44 +7,53 @@ import (
 )
 
 func TestDefaultEstimateModel(tt *testing.T) {
-	t := wrapt.WrapT(tt)
 
-	r := RunSettings{
-		OutputDir: "{{ .Name }}",
-	}
+	testId := "UNIT-RUN-003"
+	tt.Run(testId, func(tt *testing.T) {
```
seth127, line 12: I added these `tt.Run()` calls so that I could get the `testId` into the output, but I'm not 100% confident this is the right way to do it. Any input is welcome.

parsers/nmparser/parse_run_details_test.go
```diff
@@ -127,6 +129,8 @@ func TestParseRunDetails(tt *testing.T) {
 	}
 }
 
+// This was turned off Sept 2021 but
```
seth127, line 132: @elisarver I think this was you. Do you remember why you commented this out? Maybe I missed it, but I can't find any comment or note. I'm considering whether we should just delete it?

runner/consts_test.go
elisarver: It's ok to remove this. Go compiler won't compile with missing variables, and this has been commented out to no ill effect.

parsers/nmparser/parse_run_details_test.go
elisarver: I don't think so. I used Todo in all my comments.

runner/estimate_model_test.go
elisarver: It will give you the sub test name after spending a slash to the outer test function name. I think a dash works here. Spaces in a test name are concerned to underscores, btw.

parsers/nmparser/parse_run_details_test.go
kyleam: This was commented out way back in 516fc46 (wip: consume grd file; change bool to new type Status, 2019-09-19).  @elisarver just changed the commenting style in cd8ec1a (Wrapt, 2021-09-17)

seth127: ah my bad. I should've clicked deeper on the git blame. Ok, I'll delete it. 2-year old commented code shall not survive this PR.

kyleam (approved):
Going through the diff (using `--color-moved=dimmed-zebra --color-moved-ws=allow-indentation-change` to clearly mark the `tt.Run` movement), these test ID additions look good to me.  I didn't think about the ID naming for each test, but the ones I did look at more closely made sense.

Also, I checked for duplicates and didn't find any:

```
git grep 'testId :=' | cut -d= -f2  | sort | uniq -d
```

PR #267, opened 2022-06-22
README.md: update .drone.yml link

kyleam:
integration/README.md links to
https://github.com/metrumresearchgroup/bbi/blob/develop/.drone.yml

That "develop" part needs to be updated because we're going to remove
the `develop` branch and start using `main` as the base for topic
branches.  But using the full URL here is a holdover from when this
file was in the bbitest repo prior to 6efbe54 (Merge bbitest #242,
2021-09-21).  Switch to using a relative link.


seth127 (approved):
good housekeeping

Issue #194, opened 2020-10-21
clean_lvl needs more levels, and consider unifying with copy_lvl


seth127:
Currently `clean_lvl` only has two real options. `clean_lvl=0` deletes nothing (even `temp_dir` remains) and `clean_lvl=1` (or any other level) uses [this function](https://github.com/metrumresearchgroup/babylon/blob/master/cmd/nonmem.go#L545) to pull from a constant [here](https://github.com/metrumresearchgroup/babylon/blob/master/cmd/nonmem.go#L77) and add [a few more](https://github.com/metrumresearchgroup/babylon/blob/master/cmd/nonmem.go#L552) based on the run name/number.

The _intention_ however, is to have some hierarchy of importance, similar to what is defined for `copy_lvl` [here](https://github.com/metrumresearchgroup/babylon/blob/f0962aa8b239b3c863bedc031624be5ce9af6e5f/runner/consts.go#L94).

We will build a spec below for exactly what that hierarchy should be. Additionally, it would be good (primarily for maintainability in the future) if we could agree on the same hierarchy mapping for both `clean_lvl` and `copy_lvl` and unify the implementation.

seth127:
# Spec

The proposal is that a unified interface for `copy_lvl` and `clean_lvl` would have the following mapping. This is based on, though not identical to, what is [currently implemented](https://github.com/metrumresearchgroup/babylon/blob/master/runner/consts.go#L94) for `copy_lvl`.

* Level 1 deletes/copies:
  * {run}
  * {run}_ETAS
  * {run}_RMAT
  * {run}_SMAT
  * {run}.msf
  * {run}_ETAS.msf
  * {run}_RMAT.msf
  * {run}_SMAT.msf
  * all the temp stuff defined [here](https://github.com/metrumresearchgroup/babylon/blob/master/cmd/nonmem.go#L77) 
    * maybe this temp stuff is only for the delete, not the copy. We probably never want to copy all of this stuff.

* Level 2 additionally deletes/copies:
  * {run}.clt
  * {run}.coi
  * {run}.clt
  * {run}.coi
  * {run}.cpu
  * {run}.shm
  * {run}.phi

* Level 3 additionally deletes/copies:
  * {run}.xml
  * {run}.grd
  * {run}.shk
  * {run}.cor
  * {run}.cov

[...]

PR #276, opened 2022-07-28
reclean: avoid index error if no positional argument is specified

kyleam:
This small PR fixes the following error that I noticed when looking into another issue:

```
$ bbi nonmem reclean
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
[...]
```

With this PR, that becomes

```
$ bbi nonmem reclean
Error: must specify one positional argument, a directory
Usage:
  bbi nonmem reclean [flags] <directory>

Flags:
[...]
```

It also fixes the more general issue of bbi double printing errors messages for `RunE` commands (mentioned in https://github.com/metrumresearchgroup/bbi/pull/272#discussion_r932278348)

Note that this PR uses gh-272 as the base, so this shouldn't be merged before that PR.  That was done just to piggyback off of its last commit (e1e1bda) that quiets the linter (the tests from the PR trigger the same issue independently).

### DIFF
--- a/cmd/reclean.go
+++ b/cmd/reclean.go
@@ -43,7 +43,7 @@ func reclean(cmd *cobra.Command, args []string) error {
 
 func NewRecleanCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "reclean",
+		Use:   "reclean [flags] <directory>",
 		Short: "clean files in an estimation directory by clean level",
 		Long: `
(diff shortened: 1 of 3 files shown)


seth127 (approved):
Looks good. This reminds me that we should revisit [this](https://github.com/metrumresearchgroup/bbi/issues/194) at some point.

PR #277, opened 2022-07-29
drone: update to latest golangci-lint

kyleam:
Drone's lint phase is passing with no warnings, but the latest version
of golangci-lint (which I have installed locally) shows a "variable
'tokens' is only used in the if-statement" warning.  Catch drone up
with the latest release so that its results are more likely to align
with what developers see locally.


seth127 (approved):
good housekeeping

PR #279, opened 2022-08-01
nonmem: pass -maxlim=2 to nmfe by default

### LINKED ISSUE #262: Set maxlim by default
- Default value set to 2
- When maxlim is 3 we should also pass `-prcompile`; this will _force_ recompilation of some model components 
  - maxlim 3 is needed when we need more than default number of records per individual; if this is needed, it is usually in simulation context

kyleam:
This PR makes it so that `-maxlim=2` is passed to `nmfe` by default, following @kylebaron's recommendation.

Closes #262.


seth127 (approved):
This looks good. I played with it a bit, but I think, given all these small changes, I think we should make sure to give the first release candidate a thorough testing on some real world models. Probably pull @kylebaron in for that, and also run some of the more complicated models that we've seen give us trouble (specifically the one that was failing with the defaults before and hopefully will pass with `maxlim=2`).

PR #281, opened 2022-08-03
summary: return error if lst file lacks DATA line

kyleam:
This series addresses "weird result 2" that @seth127 reported at https://github.com/metrumresearchgroup/bbr/pull/527#pullrequestreview-1058066983

The last commit is the main one of interest.  It adds a check that relies on any valid lst file having a `$DATA` line.  **Is that a safe assumption?**

The first commit isn't conceptually related, but I noticed it while working on this and it touches the same test ref output, so I've included it here.


seth127 (approved):
Looks good to me. I tried it out [here](https://github.com/metrumresearchgroup/bbr/pull/527#pullrequestreview-1061057228) and it works as expected. Thanks.

Issue #269, opened 2022-06-30
Automatically clean up FDATA.csv whenever FDATA is cleaned up


kylebaron:
# Summary 
As a user, I want bbi to treat `FDATA.csv` the same way it treats `FDATA` when cleaning up files post-run

> With NONMEM 7.5, an additional file, FDATA.csv is  produced  that
outputs  the contents of its input data file (typically FDATA) in
a comma delimited file format, so you can check how NONMEM inter-
prets  the  input data.

kyleam:
Thanks @kylebaron

The core code change may just be (untested):

```diff
diff --git a/cmd/nonmem.go b/cmd/nonmem.go
index a416650..249adca 100644
--- a/cmd/nonmem.go
+++ b/cmd/nonmem.go
@@ -76,6 +76,7 @@ var nonMemTemporaryFiles []string = []string{
 	"compile.lnk",
 	"FCON",
 	"FDATA",
+	"FDATA.csv",
 	"FMSG",
 	"FREPORT",
 	"FSIZES",
diff --git a/runner/consts.go b/runner/consts.go
index edd1e52..166690b 100644
--- a/runner/consts.go
+++ b/runner/consts.go
@@ -14,6 +14,7 @@ func EstOutputFileCleanLevels(r string) map[string]int {
 	EstOutputFiles["compile.lnk"] = 1
 	EstOutputFiles["FCON"] = 1
 	EstOutputFiles["FDATA"] = 1
+	EstOutputFiles["FDATA.csv"] = 1
 	EstOutputFiles["FMSG"] = 1
 	EstOutputFiles["FREPORT"] = 1
 	EstOutputFiles["FSIZES"] = 1
```

I haven't looked closely at the testing situation yet, but an additional aspect is that right now drone is only testing with `NMVERSION: nm74gf`

kylebaron:
@kyleam Seems reasonable; I searched the code a bit to make sure this is where stuff was happening and ran across this too, but don't know the code enough (at all) to know if it would be any other place. But I feel like following whatever we do with `FDATA`  and clean level should work.  

Good point on NONMEM version; should only see this appearing with 75.  Would be good investment to switch to nm75 as lots of new stuff there and most projects should be using this version going forward.

dpastoor:
Yep that should do it

PR #282, opened 2022-08-04
Clean up NONMEM 7.5's FDATA.csv

kyleam:
This series addresses gh-269 by
 * updating the relevant spots to include FDATA.csv
 * switching Drone over to testing with NONMEM 7.5 (except for the NMQUAL tests, which still need to use 74gf)
 * sprinkling various tests with an assertion that FDATA and FDATA.csv are cleaned up

---

@seth127 Any opinions on where, if anywhere, this should be added to stories/reqs?


integration/nonmem.go
```diff
@@ -50,6 +50,19 @@ func AssertNonMemCreatedOutputFiles(t *wrapt.T, details NonMemTestingDetails) {
 	}
 }
 
+func AssertNonMemCleanedUpFiles(t *wrapt.T, details NonMemTestingDetails) {
+	t.Helper()
+
+	expectMissing := []string{
```
seth127, line 56: I guess maybe consider whether we want to check for any other cleaned up files in this function? I don't have a strong feeling about that, but it does seem a little arbitrary to to just check these two. 

On the other hand, if you think it doesn't add much value and might get brittle to add more then I'm fine with leaving it.

seth127 (approved):
One minor comment, but feel free to ignore. This looks good to me.

I don't think we need any changes in the requirements or stories for this.

integration/nonmem.go
kyleam: > it does seem a little arbitrary to to just check these two.

Fair enough.  I've pushed a change that now checks all of the level 1 files returned by `EstOutputFileCleanLevels`.

Locally I verified that it still fails when the relevant part of c062ae94 (nonmem: clean up FDATA.csv) is reverted.

<details>
<summary>regression check</summary>

```
ip-10-128-42-167:bbi$ ( make && cd integration && go test -run TestBbiCompletesLocalExecution )
cd cmd/bbi; go install -ldflags "-X main.buildTime=`date +%FT%T%z`"
PASS
ok      github.com/metrumresearchgroup/bbi/integration  6.583s

ip-10-128-42-167:bbi$ git diff
diff --git a/cmd/nonmem.go b/cmd/nonmem.go
index 43294bd..07082b7 100644
--- a/cmd/nonmem.go
+++ b/cmd/nonmem.go
@@ -76,7 +76,6 @@ var nonMemTemporaryFiles []string = []string{
        "compile.lnk",
        "FCON",
        "FDATA",
-       "FDATA.csv",
        "FMSG",
        "FREPORT",
        "FSIZES",

ip-10-128-42-167:bbi$ ( make && cd integration && go test -run TestBbiCompletesLocalExecution )
cd cmd/bbi; go install -ldflags "-X main.buildTime=`date +%FT%T%z`"
--- FAIL: TestBbiCompletesLocalExecution (6.41s)
    --- FAIL: TestBbiCompletesLocalExecution/acop[INT-LOCAL-001] (6.40s)
        bbi_local_test.go:32: Beginning work with /tmp/bbi-test-LRrS3w9/working as the root
        --- FAIL: TestBbiCompletesLocalExecution/acop[INT-LOCAL-001]/acop (6.38s)
            bbi_local_test.go:65:
                        Error Trace:    nonmem.go:60
                                                                bbi_local_test.go:65
                                                                wrapt.go:79
                        Error:          file "/tmp/bbi-test-LRrS3w9/working/acop/acop/FDATA.csv" exists
                        Test:           TestBbiCompletesLocalExecution/acop[INT-LOCAL-001]/acop
FAIL
exit status 1
FAIL    github.com/metrumresearchgroup/bbi/integration  6.417s
```

</details>

Issue #201, opened 2020-12-21
summary parses $PROBLEM incorrectly


seth127:
[Original post](https://github.com/metrumresearchgroup/rbabylon/issues/296) from @callistosp on the rbabylon repo:

## Problem:
The model summary object contains information parsed from the model file stored in `run_details`. One of the variables under `run_details` is `$problem_text`, which appears to parse the $PROBLEM statement from the control stream.

The code currently parses the $PROBLEM statement incorrectly. "LEM" is appended to this value in a subset of models. I believe this is due to the regular expression parser grabs anything after "$PROB" from the control stream. However, users can use either $PROB or $PROBLEM for this chunk of code.

callistosp:
oops, sorry @seth127 didn't realize this was a bbi issue

seth127:
This has been lingering way too long. We just need to knock it out.

callistosp:
this has been around so long I forgot I filed this issue

PR #283, opened 2022-08-04
nmparser: handle $PROBLEM in addition to $PROB

kyleam:
This PR fixes the `$PROBLEM` parsing issue described in gh-201.  The last commit is the main one of interest.

Note that there are lots of file changes, but most of those are just golden files where "LEM"s are being dropped:

```
$ git diff --name-only main.. ':!*golden*'
parsers/nmparser/parse_run_details.go
parsers/nmparser/parse_run_details_test.go
```


kyleam:
Bah, I forgot the req.  Will amend last commit.

seth127 (approved):
Great stuff here.

PR #287, opened 2022-08-19
validation: requirements cleanup

kyleam:
Prune unlinked requirements, and eliminate requirement ID gaps.

### DIFF
--- a/validation/requirements.yaml
+++ b/validation/requirements.yaml
@@ -2,10 +2,6 @@ CFG-R001:
   description: BBI config JSON gets created
   tests:
   - INT-CFG-001
-CFG-R002:
-  description: Config values are correct in written file
-  tests:
-  - INT-CFG-002
 COVCOR-R001:
   description: CovCor parses .cov and .cor files
   tests:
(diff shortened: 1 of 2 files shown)


barrettk:
Just to get some additional context, was there a reason you chose to remove all those requirements vs. adding them to new stories? Its tough to know from an outside perspective, but most look like they're related to helper functions/might not be important enough to include, but just wanted to be sure.

kyleam:
> was there a reason you chose to remove all those requirements vs. adding them to new stories?

To expand on what I said in the commit message...

```
validation: prune unlinked requirements

The previous release didn't have requirements (stories were linked
directly to tests).  A script generated requirements.yaml based on the
tests, but that includes a lot of requirements that aren't actually
linked to a story (because the corresponding test ID wasn't either).

Clean these up.
```


... these extra ones are just artifacts of the import process.  We are retaining all the stories that are currently in the validation docs and adding one more.

So, perhaps someone will decide we need to flesh out bbi's stories, but 1) I'm not pushing for that and 2) the decision to move from `tests -> stories` to `tests -> requirements -> stories` does not imply that we need to change the current state of bbi's stories.

barrettk:
Yeah ok that makes sense. Thanks for the additional context

barrettk (approved):
Looks good

kyleam:
@barrettk Thanks for the review.  Not the most exciting thing to look through :)

PR #289, opened 2022-09-07
summary: account for chain method in incomplete run check

### LINKED ISSUE #288: METHOD=CHAIN and METHOD=BAYES together
# Summary

Ref: https://github.com/metrumresearchgroup/bbi/pull/272/files#r965087052

When we run Bayesian estimation, we use two `$ESTIMATION` records

```
$EST METHOD = CHAIN ...
$EST METHOD = BAYES ...
```

The `METHOD=CHAIN` record does not produce any parameter estimates.

So you will find two estimation methods in the `.lst` file (when looking for `#METH`), but there will only be one set of estimates. 

This setup currently produces an error: 

```r
 - Installed Version: 3.2.0  - Current release: 3.2.0
> x <- read_model(here("model/nonmem/pk/7/7_1"))
> y <- model_summary(x)
Error: nonmem_summary('/data/home/kyleb/project.mrg/bi/dinamo/model/nonmem/pk/7/7_1') failed with the following error. This may be because the modeling run has not finished successfully.

ERROR:
Error in check_status_code(p$get_exit_status(), output, .cmd_args):

`bbi nonmem summary 7_1 --json` returned status code 1 -- STDOUT and STDERR:
{
"run_details": {
"postprocess_time": 0,
"cpu_time": 0
},
"run_heuristics": {
"covariance_step_aborted": false,
"large_condition_number": false,
"eigenvalue_issues": false,
"correlations_not_ok": false,
"parameter_near_boundary": false,
"hessian_reset": false,
"has_final_zero_gradient": false,
"minimization_terminated": false,
"eta_pval_significant": false,
"prderr": false
},
"parameter_names": {},
"success": false,
"error_msg": "lengths of parameter data (1) and estimation methods (2) differ. Incomplete run?"
}
```

kyleam:
As of the last release (specifically 7c0caa76), GetModelOutput() compares the lengths of RunDetails.EstimationMethods and of the ParseParamsExt() return value, and errors if they don't match.  The intent is to catch incomplete runs, but this check is incorrectly triggered for METHOD=CHAIN because that method does not have corresponding estimation data.

Update the check so that the chain method is not counted toward the expected length.  Based on the NONMEM help, it appears that all other method types should have corresponding estimation values.

Closes #288.

---

I don't think any adjustments are needed on the bbr side.  Using the same model from which the test data here was derived, I confirmed that, following this bbi change, I can call `model_summary` with no errors.

cc @kylebaron


parsers/nmparser/get_model_output.go
```diff
@@ -111,11 +123,11 @@ func GetModelOutput(lstPath string, ext ModelOutputFile, grd bool, shk bool) (Su
...
 		extData, parameterNames := ParseParamsExt(extLinesParsed)
 		lenData := len(extData)
-		lenMethods := len(results.RunDetails.EstimationMethods)
-		if lenData != lenMethods {
+		lenExpected := expectedDataLen(results.RunDetails.EstimationMethods)
+		if lenData != lenExpected {
 			return results, fmt.Errorf(
-				"lengths of parameter data (%d) and estimation methods (%d) differ. Incomplete run?",
-				lenData, lenMethods)
+				"lengths of parameter data (%d) non-chain estimation methods (%d) differ. Incomplete run?",
```
kyleam, line 129: Doh, accidentally swallowed an "and" there.

kyleam:
Drone has some failures. 

https://github-drone.metrumrg.com/metrumresearchgroup/bbi/954/2/6

seth127 (approved):
Looks good. Thanks for catching this @kylebaron

PR #297, opened 2023-02-15
run: quote tokens if needed when generating shell scripts

kyleam:
This PR was prompted by the following case on Windows:

 * `bbi init` works on Windows as of e09a001 (init: make compatible with Windows, 2023-02-02).  Unsurprisingly it writes the `nonmem.*.home` values in the yaml with "\\"

 * The bash script generated by `bbi nonmem run local` doesn't quote that path, leading to an error.

That points to a more general problem of bbi not quoting arbitrary paths and file names in the bash scripts it generates.  This PR adds a helper function to quote values if needed and uses it in the spots that generate bash scripts.   More details are in the commit message, but there are still cases that don't work related to the path handling by nmfe or nmqual/autolog.pl and to SGE issues I don't fully understand.  The goal at this point is just to make some cases work while not breaking any cases that worked before.

---

This PR should address the Windows case, though we're still working setting up a Windows instance with NONMEM. 
 Here's my toy example on Windows that indicates to me that this quoting should work:

```
Administrator@EC2AMAZ-V9JAO7M MINGW64 ~/scratch
$ cat run-a-bat.sh
#!/bin/sh

'C:\Users\Administrator\scratch\a.bat'
C:\\Users\\Administrator\\scratch\\a.bat
C:/Users/Administrator/scratch/a.bat
/c/Users/Administrator/scratch/a.bat


Administrator@EC2AMAZ-V9JAO7M MINGW64 ~/scratch
$ cat a.bat
ECHO BAAAAAAATT

Administrator@EC2AMAZ-V9JAO7M MINGW64 ~/scratch
$ ./run-a-bat.sh

C:\Users\Administrator\scratch>ECHO BAAAAAAATT
BAAAAAAATT

C:\Users\Administrator\scratch>ECHO BAAAAAAATT
BAAAAAAATT

C:\Users\Administrator\scratch>ECHO BAAAAAAATT
BAAAAAAATT

C:\Users\Administrator\scratch>ECHO BAAAAAAATT
BAAAAAAATT
```


seth127 (approved):
Looks good to me. Tests cover obvious cases.

PR #299, opened 2023-02-28
summary: include empty problem_text in JSON

kyleam:
The RunDetails struct marks ProblemText with omitempty, so the empty value is dropped from the JSON output.  Dropping problem_text causes downstream issues with bbr, which requires the key when creating summary log objects.

Drop the "omitempty" tag so that the problem_text is always included.

Note that what problem specifications got parsed as an empty string changed in v3.2.0:

 * Before c0aff14 (nmparser: handle $PROBLEM in addition to $PROB, 2022-08-04), "$PROB" or "$PROB " would get extracted to an empty string.  On the other hand, "$PROBLEM" got incorrectly parsed to "LEM", so it never led to an empty string.

 * After c0aff14, "$PROB " and "$PROBLEM " (note trailing space) get extracted to the empty string.  Without the space, the value remains at DefaultString and included in the JSON.


seth127 (approved):
looks good, thanks

PR #304, opened 2023-05-23
init: warn on ID collisions and support NONMEM directories with periods

kyleam:
This series was prompted by `bbi init` mishandling a directory with periods.  For example, a directory with the name "7.5.0" will lead to this entry:

```
nonmem:
  "7": {}
```

<details>
<summary>demo script</summary>

```sh
#!/bin/sh

set -eu

tdir=$(mktemp -d "${TMPDIR:-/tmp}"/bbi-init-test-XXXXXXX)

test $# != 0 || {
    echo "usage: $0 <NONMEM name>"
    exit 1
}

cd "$tdir"

for nm_name in "$@"
do

    for sdir in license run util source
    do
        mkdir -p d/"$nm_name"/"$sdir"
    done

    :> d/"$nm_name"/license/nonmem.lic
    :> d/"$nm_name"/run/nmfe75
    chmod +x d/"$nm_name"/run/nmfe75
done

bbi init --dir d

cat bbi.yaml
```

```
./demo.sh 7.5.0 | grep nonmem -A1
nonmem:
  "7": {}
```

</details>

After the last commit, that will be accessible as "7-5-0".

---

```
init: rename variable for style consistency
init_test: check that error is non-nil before inspecting message
init: convert a panic to a returned error
```

These first three commits touch up minor things in the area that aren't directly related to the original problem.

```
init: extract logic for finding installation directories to function
init: generate NonMemDetail map directly
```

The next commits split up and reworks the implementation a bit, mostly for easier testing (but I also think it makes things a bit clearer).

```
init: warn on identifier collisions
init: detect collision due case insensitive keys
```

These two commits add collision warnings for two cases that could be hit before this PR.

```
init: sanitize config identifier for nonmem entries
```

The last commit handles the "." to "-" replacement, warning if that creates any collisions.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -265,7 +265,8 @@ func hasNMQual(path string) bool {
 func makeIdentifiers(locs []string) map[string]string {
 	ids := make(map[string]string)
 	for _, loc := range locs {
-		id := filepath.Base(loc)
+		// Viper keys are case insensitive.
+		id := strings.ToLower(filepath.Base(loc))
 		if loc1, exists := ids[id]; exists {
 			log.Warnf("Ignoring %q because its key collides with %q\n", loc1, loc)
 
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -27,39 +27,16 @@ func initializer(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("get dir string: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


seth127 (approved):
Lookin good. Thanks for fixing this.

Issue #285, opened 2022-08-11
automatically remove parallelization files


callistosp:
When running a NONMEM model in parallel, many folders with the prefix WK_ are produced. It would be nice if bbi automatically removed these files. There are also files produced after every run with the extension `.o#` and `.po#` where the # is the number assigned by the grid. I think these could also be automatically handled by bbi to prevent buildup of junk files in version control when saving a whole model directory.

shairozan:
Would also probably make sense to update the .gitignore file so that even if the files get left around they wouldn't be committed in a git environment

jacobleander:
Picking up this thread again. I would also prefer having these files deleted as part of the clean up of the model directory. IF there are users who would really like these files to not be deleted, could there be a global option in bbi where this can be turned on or off (default setting should then be that the files are removed). Thoughts @seth127 ?

seth127:
Thanks for getting this going again @jacobleander. I think this is the right idea:

> could there be a global option in bbi where this can be turned on or off (default setting should then be that the files are removed)

We currently have a `--clean_lvl` flag that controls our deletion of other NONMEM temp files. I would think we could use that to control this behavior as well. Any thoughts on that @kyleam ?

seth127:
Also, @kyleam if we decide to do this through `--clean_lvl`, we should consider revisiting this old issue while we're in there: https://github.com/metrumresearchgroup/bbi/issues/194

kyleam:
I've put up a draft at gh-306.  There's an issue with the CI that needs to be sorted out before considering merging that end.

@seth127:
> We currently have a `--clean_lvl` flag that controls our deletion of other NONMEM temp files. I would think we could use that to control this behavior as well. Any thoughts on that @kyleam ?

Conceptually I think that makes sense, but there's already [a spot](https://github.com/metrumresearchgroup/bbi/blob/8c65e319a1a8e4fa0b213903c22870537c95367b/cmd/nonmem.go#L498-L515) in `filesToCleanup` that takes care of cleaning up some parallel file related things.  It doesn't consider the clean level, so in gh-306 I decided to just follow that.  I suppose we could update that whole block to be skipped when `clean_level=0`, but I think in practice unconditionally cleaning empty `WK_` files is probably fine, at least until we take on something like the more comprehensive rework proposed in gh-194.
Let me know if you think the `clean_level` guard is worth doing at this point.


---


@callistosp:
[...]

PR #306, opened 2023-09-22
Clean up WK_ files

kyleam:
Work towards gh-285.

The base is currently [failing on Drone](https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1021), so that needs to be resolved before moving forward with this.


kyleam:
This is ready to review but should be merged after its base (gh-307).  See the message of the second commit for details about these `WK_*` files.

seth127 (approved):
This approach makes sense to me. Thanks for the detailed commit messages.

PR #307, opened 2023-10-06
integration: delete now failing MTP models

kyleam:
Tests that use integration/testdata/240/Executable_MTP.mod or models derived from it have started to fail on Drone, apparently due to the NONMEM 7.5.0 to 7.5.1 update.  The tests pass under NM_VERSION=nm74gf.

The test failures are triggered by missing grd files.  These runs have the following error:

    0PRED EXIT CODE = 1
    [...]
     OCCURS DURING SEARCH FOR ETA AT INITIAL VALUE, ETA=0
     ERROR IN LSODA: CODE  -3

The -3 LSODA error "means illegal input was detected", but, even with SME input, I've been unable to tweak the model to resolve the issue.

The details of this model aren't relevant for the tests.  Resolve the failures by no longer using this model.  Derive the model variations from the acop model.


seth127 (approved):
Makes sense. Thanks for looking into this.

> The details of this model aren't relevant for the tests.  Resolve the
failures by no longer using this model.

That makes sense to me.

PR #308, opened 2023-10-06
integration: test period_test scenario

kyleam:
As of aa686d3 (tests: removing more scenarios from tests that don't need so many divergent paths [...], 2021-09-23), the period_test scenario is no longer used by any tests.  This scenario provides a regression test for a fix in v2.3.1, so it's worth keeping around.

Restore its use in TestBbiCompletesLocalExecution.

### DIFF
--- a/integration/bbi_local_test.go
+++ b/integration/bbi_local_test.go
@@ -24,6 +24,7 @@ func TestBbiCompletesLocalExecution(tt *testing.T) {
 		{name: "metrum_std"},
 		{name: "leading-path-with space"},
 		{name: "modquote"},
+		{name: "period_test"},
 		{name: "datacomment"},
 	}
 


kyleam:
This is ready to review but should be merged after its base (gh-307).

kyleam:
The [integration][1] [tests][2] for this branch took about 20 minutes.  The [base][3] run took around 12 minutes, which makes me worried this is adding a bit of time.  However...

 * I don't expect this to add much time because 1) it's just adding one _local_ test and 2) it only adds ~5 seconds on my end.

 * other runs without this commit took longer than the base's run (~14-16 minutes)

   https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1037/2/6
   https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1036/2/6
   https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1039/2/6
   https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1038/2/6

 * a re-push of the same tree as this PR took ~16 minutes

   https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1043/2/6

So my current guess is that the timing variability above is due to something other than this PR's additional scenario (e.g., the runner being primed for SGE).

[1]: https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1034/2/6
[2]: https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1035/2/6
[3]: https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1031/2/6

seth127 (approved):
Adding this back is a good call, though the long-running tests are a bit concerning.

> I don't expect this to add much time because 1) it's just adding one local test and 2) it only adds ~5 seconds on my end.

This ^ makes sense to me, and I had the same thought when I looked at it.

> So my current guess is that the timing variability above is due to something other than this PR's additional scenario (e.g., the runner being primed for SGE).

I was thinking the same thing. Our test suite time is extremely variable: [this run](https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1044/2/6) from #310 only took 9 minutes, while essentially the same code took 25 minutes in the [companion build](https://github-drone.metrumrg.com/metrumresearchgroup/bbi/1041/2/1). 

That is probably worth investigating further at some point, and in particular making sure that this 8b64c93373bf9c512119fec23b6d126555bb9219 commit didn't _actually_ increase the run time significantly, but I don't think it needs to hold up this PR. Mostly because both of us have good reason to suspect it's unrelated. I'll leave it to your discretion whether that is worth opening an issue about, or if we should just track it elsewhere.

kyleam:
> I was thinking the same thing [...]

Thanks for your input here (and good point about the difference between the companion build).

PR #323, opened 2024-08-09
ci: switch to GitHub Actions

kyleam:
The Drone tests for the latest push failed to launch.  Given that we have been planning to switch to GitHub Actions, it's not worth it to troubleshoot what's going on.  So, this PR does the switch.  As discussed offline a few times, the strategy is to go with the GitHub-hosted runners and run all the tests that don't require NONMEM, leaving it to us to execute the full test suite outside of CI (which, at the very least, we'll be doing as part of validation anyway).

There are two prep commits:

 * The first fixes an issue flagged by the new golang-lint workflow.  (We executed that before, but not with the minimum version of Go specified in `go.mod`.)  
 * The second splits the integration tests into two packages so that we can more easily test the things that do not require NONMEM (e.g., summary tests).

### Confirmation that integration/nonmem tests are passing

<details>
<summary>script</summary>

```sh
#!/bin/sh

set -eu

test -n "${METWORX_VERSION-}" || {
    printf >&2 'Metworx is required to run integration tests\n'
    exit 1
}

root=${BBI_TEST_ROOT-/data/bbi-tests}
mkdir -p "$root"
tdir=$(mktemp -d "$root"/run-XXXXX)
trap 'rm -rf "$tdir"' 0

BBI_GRID_NAME_PREFIX=$(basename "$tdir")
export BBI_GRID_NAME_PREFIX
export LOCAL=true
export MPIEXEC_PATH=/usr/bin/mpiexec
export NMQUAL=false
export NMVERSION=nm75
export NMVERSION_NMQUAL=nm74gf
export NONMEMROOT=/opt/NONMEM
export POST_EXECUTION=true
export ROOT_EXECUTION_DIR="$tdir"
export SGE=true
export SGE_ARCH=lx-amd64
export SGE_CELL=default
export SGE_CLUSTER_NAME=p6444
export SGE_EXECD_PORT=6445
export SGE_QMASTER_PORT=6444
export SGE_ROOT=/opt/sge

bin=$tdir/bin
mkdir "$bin"

version=$(git describe --always --dirty)

# shellcheck disable=SC2086
go build \
   -ldflags "-X github.com/metrumresearchgroup/bbi/cmd.VERSION=$version" \
   -o "$bin/bbi" cmd/bbi/main.go

export PATH="$bin:$PATH"

printf >&2 'bbi path: %s\nbbi version: %s\ntree: %s\n' \
       "$(command -v bbi)" "$(bbi version)" "$(git rev-parse 'HEAD^{tree}')"

cd integration
bbi init --dir /opt/NONMEM --threads 2

go test -count 1 "$@" ./postrun ./nonmem
```

</details>

```
$ ./t.sh
bbi path: /data/bbi-tests/run-zIsYL/bin/bbi
bbi version: v3.3.0-3-g97dca5c
tree: ee97656114c965f16c7f17fd7f375994f0db143b
ok  	github.com/metrumresearchgroup/bbi/integration/postrun	0.469s
ok  	github.com/metrumresearchgroup/bbi/integration/nonmem	455.006s
```


.github/workflows/main.yaml
```diff
@@ -0,0 +1,67 @@
...
+          go build -o "$bin/bbi" \
+            -ldflags "-X github.com/metrumresearchgroup/bbi/cmd.VERSION=$version" \
+            cmd/bbi/main.go
+          echo "$bin" >>$GITHUB_PATH
+      - name: Integration tests (postrun)
+        shell: bash
+        run: |
+          bbi version
+          go test ./integration/postrun
+  release:
```
kyleam, line 45: This hasn't been tested.  My thinking is that we can do a prerelease after the validation rework lands, and I can adjust as needed then.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          bin=$(mktemp -d)
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+            sh -s -- -b "$bin" v1.47.2
+          echo "$bin" >>$GITHUB_PATH
+      - name: Run golangci-lint
+        shell: bash
+        run: |
+          command -v golangci-lint
+          golangci-lint run --out-format=github-actions
```
kyleam, line 33: Here's an example of how this will fail when it flags an issue: https://github.com/metrumresearchgroup/bbi/actions/runs/10321533479

seth127 (approved):
This all looks good to me. I agree with the approach of getting this running and then testing the release step later (with a pre-release).

> run all the tests that don't require NONMEM, leaving it to us to execute the full test suite outside of CI (which, at the very least, we'll be doing as part of validation anyway)

Yes, we discussed this and I still think this is the right path forward. Tagging @shairozan here, because this came up in a conversation that we had earlier this week.

PR #324, opened 2024-08-13
ci: update lint workflow

kyleam:
* The incoming valtools branch triggers the lint workflow to fail for reasons related to a dot import.

   https://github.com/metrumresearchgroup/bbi/actions/runs/10370260543/job/28707761437

   I don't understand why the lint workflow is happy on main but not after that series, but, either way, it's a good idea to use a qualified import, so I just switched to that.

 * After that, the lint workflow fails with an obscure error I don't understand:

   https://github.com/metrumresearchgroup/bbi/actions/runs/10370823460/job/28709603032
  
   Resolve that by upgrading the lint workflow the latest golangci-lint.  To do that, we need to 1) update the golangci configuration and 2) reformat doc comments in the code base with a newer Go.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,11 +6,6 @@ on:
       - 'scratch/**'
   pull_request:
 
-# Note: this avoids golangci/golangci-lint-action to work around an
-# error triggered by us pinning to a golangci-lint version built with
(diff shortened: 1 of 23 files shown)


seth127 (approved):
Makes sense to me, and CI looks clean. Thanks.

PR #334, opened 2024-10-08
nmparser: guard against NaN for all incoming floats

kyleam:
@callistosp reported a case where `bbi nonmem summary --json` fails to encode the result:

```
FATA[0000] json: unsupported value: NaN
```

To avoid this and related errors, this PR makes all spots that parse floats replace `NaN` with `DefaultFloat64`.


integration/nmless/testdata/bbi_summary/aa_golden_files/acop-nan.golden.json
```diff
@@ -49,12 +49,12 @@
...
 					2.31034,
 					54.9596,
-					464.659,
+					-999999999,
 					-0.0805722,
 					4.1303
 				],
 				"omega": [
-					0.0964407,
+					-999999999,
```
seth127, line 57: I must be missing something, but I'm unclear why these golden files are changing to now included NaN's (or, more specifically `-999` aka `DefaultFloat64`). I understand the motivation, but I don't see where we change any test files to now expect NaN in these spots. Thanks in advance for explaining.

kyleam: Sorry, I should have explained that somewhere.  I tweaked the `integration/nmless/testdata/bbi_summary/acop-nan/acop-nan.ext` input to introduce NaNs in a bunch of spots.  That was an existing test case for related to NaNs, but only set the objective function values to NaN in the lst file (75d8f5c6, 2022-06-23).

So the golden file changes are coming from me adjusting the input.

kyleam: I've updated the commit message to include this: 

> For testing, piggyback on top of summary's existing acop-nan test case by setting many values in its ext file to NaN.

seth127: That makes sense. Thanks for the explanation.

seth127 (approved):

PR #335, opened 2024-10-09
treewide: drop inappropriate executable bits

### LINKED ISSUE #181: babylon.yaml should not have executable bit set
`bbi init` creates `babylon.yaml` with the executable bit set:

```sh
➜  /tmp /data/apps2/bbi init
➜  /tmp ls -l babylon.yaml
-rwxr-xr-x 1 seanw seanw 261 Aug 28 13:01 babylon.yaml
```

This file should not be marked as executable. Perhaps we could instead use mode `644`.

kyleam:
Adjust various spots that set the executable bit on files that cannot be executed (e.g., bbi.yaml and {model}.out files).

Close #181.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -44,7 +44,7 @@ func initializer(cmd *cobra.Command, _ []string) error {
 	}
 
 	// Write the byte array to file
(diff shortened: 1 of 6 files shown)


seth127 (approved):
Good housekeeping. @copernican should be pleased.

PR #338, opened 2024-12-06
Support Slurm execution

kyleam:
This series adds Slurm support via a new `bbi nonmem run slurm` subcommand.

### Overview

```
[01] cmd: move errpanic from sge to root
[02] cmd: remove unnecessary appends
[03] cmd: move template error checks immediately after call sites
[04] cmd: drop extra spaces from nmfe command string
[05] sge: simplify an append
```

These are cleanup commits that prepare for pulling out code from sge.go that will also be used for Slurm.

```
[06] local, sge: decouple script templates
[07] local, sge: do shell quoting at template level
[08] sge: convert qsub options to template directives
```

These commits move the qsub options to the template, so that submission command is reduced to `qsub {submission script}` and switching to submission command to Slurm can be done by swapping "qsub" for "sbatch".

```
[09] sge: simplify the description of --grid_name_prefix
[10] sge: improve the description of --bbi_binary
[11] sge: drop sgeOperation struct
[12] sge: use consistent parameter name for SGEModel methods
```

More cleanup in preparation for splitting `sge.go` into shared code and SGE-specific bits.

```
[13] sge: decouple SGE-specific details from grid submission
```

This commit does the main work of the series, extracting the generic "submit to grid" code from `sge.go` to `grid.go`.

```
[14] run-integration-tests: adjust MPIEXEC_PATH for Metworx compatibility
[15] integration/nonmem: prune commented code in SGE tests
[16] integration/nonmem: drop test IDs from SGE tests
[17] integration/nonmem: extract SGE checks to test helpers
[18] vt-test: allow skipped tests
```

These commits prepare the integration tests for Slurm.

```
[19] nonmem run: support Slurm submission
```

This commit wires up and exposes `bbi nonmem run slurm`.

```
[20] slurm: include modquote test case
```

Adds a test case that works for Slurm but not for SGE.


kyleam:
`make vt-test` is passing for me on both SGE and Slurm workflows, with the caveat that both the SGE and Slurm runs fail (inconsistently) if a worker isn't primed ([related discussion](https://github.com/metrumresearchgroup/bbi/pull/325#issuecomment-2289299368)).  I'm still not sure what's going on there, though I'm fairly convinced that the wait is in effect and that it's not a matter of increasing the wait time.

cmd/grid.go
```diff
@@ -0,0 +1,392 @@
...
+	IgnoreError func(error, string) bool
+}
+
+// gridModel is the struct used for grid operations.
+type gridModel struct {
+	Spec                 *gridSpec
+	Nonmem               *NonMemModel
+	Cancel               chan bool
+	postworkInstructions *PostExecutionHookEnvironment
+}
```
kyleam, line 64: After this point, the code in this file is coming with light modifications from what was in `sge.go`.  Reviewing the diff locally will likely be much easier to digest.

Similarly, most of the `integration/nonmem/grid.go` content is coming unchanged from `bbi_sge_test.go`.

kyleam: @shairozan Good point, I agree it makes sense to move grid.go from under `cmd/`.

kyleam: Taking a quick look, I think that restructuring would require a larger rework because `grid.go` depends on other things defined in `cmd/` (e.g., `NonMemModel` defined `cmd/nonmem.go` and `PostExecutionHookEnvironment` defined in `cmd/run.go`).  In the long run, I don't think that larger rework would be a bad thing, though for now I'm okay with introducing a `cmd/` file that doesn't actually map to a CLI subcommand.  (We could of course absorb the code into an existing file like `cmd/run.go`, though in my view that's not an improvement organization-wise.)

seth127 (approved):
This all looks great to me, nice work. 

I did some manual testing (using the `bbr` in https://github.com/metrumresearchgroup/bbr/pull/724) on both SGE and Slurm and all worked as I expected. At a high-level, my manual tests used the [106](https://github.com/metrumresearchgroup/expo1-nonmem-foce/blob/main/model/pk/106.ctl) and [200](https://github.com/metrumresearchgroup/expo1-nonmem-foce/blob/main/model/pk-parallel/200.ctl) models from our Expo and covered the following (with both SGE and Slurm):

* submit locally, both single-threaded and parallel=TRUE (threads=8)
* submit to the grid, both single-threaded and parallel=TRUE (threads=8)
* run a bootstrap of 200 replicates, submitted to the grid in batches of 100 (via [bbr bootstrap](https://metrumresearchgroup.github.io/bbr/reference/setup_bootstrap_run.html))

cmd/slurm.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/spf13/viper"
+)
+
+const slurmTemplate string = `#!/bin/bash
+#SBATCH --job-name={{.JobName | shquote}}
+#SBATCH --output=slurm.out
+#SBATCH --export=ALL
+{{- if .Config.Parallel}}
+#SBATCH --ntasks={{.Config.Threads}}{{end}}
+#SBATCH --chdir={{.WorkingDirectory | shquote}}
```
kyleam, line 16: Outside the context of Metworx, I suspect callers may end up needing to set additional parameters.  I'm not sure exactly how that should look (and am not proposing it for this PR), but, as mentioned in 4e7df052 (sge: convert qsub options to template directives, 2024-12-06), I think this series gets us closer to making that happen:

> And now that submission happens with a plain 'qsub {script}' call, it
> opens up the possibility of allowing callers to override the template
> (e.g., via a command-line argument).

PR #339, opened 2024-12-16
nonmem: fix relay of --licfile option to nmfe

### LINKED ISSUE #337: Passthrough for nmfe licfile seems to not work
I am using version 3.3.0. I'm testing our containerized setup for bbi and found I needed to provide the NONMEM license file. Configuring through `nmfe_options->license_file`  was fine, but `--licfile` as in `bbi nonmem --licfile=(...) run ...` did not work. Other passthrough constructs like `--background` work fine (and update the run's local bbi.yaml), so I suspect this is related to `licfile` being the command line flag and `license_file` being the config item label.

kyleam:
Specifying --licfile does not relay the -licfile to the nmfe call in the {model}.sh script because there is a mismatch between the command-line option name (licfile) and the configuration (license_file).

Adjust the viper lookup to use the correct identifier.

Fixes #337


seth127 (approved):
Good clean fix, thanks

PR #343, opened 2025-06-05
site: include json file with latest non-dev tag

kyleam:
Provide bbr with a way to get the latest bbi version without relying on the GitHub API in order to sidestep failures due to rate limiting.

Re: https://github.com/metrumresearchgroup/bbr/issues/736

---

Deployed via push to a scratch branch:
https://github.com/metrumresearchgroup/bbi/actions/runs/15471070684/job/43555331265

```
$ curl -fSsL https://metrumresearchgroup.github.io/bbi/latest.json
{"version": "v3.4.0"}
```

### DIFF
--- a/.github/workflows/site.yaml
+++ b/.github/workflows/site.yaml
@@ -3,8 +3,9 @@ on:
   push:
     branches:
       - scratch/deploy-site
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+  release:
+    types:
+      - released
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


andersone1:
@kyleam 

Does the word "release" on line 6 of site.yaml need to be intended **_indented_*** so its at the same level as "push"?
image

Edit: I am looking here:

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

kyleam:
> Does the word "release" on line 6 [...]

Doh, thank you.

andersone1 (approved):

